PR #1, opened 2023-07-26
Initial commit

antarcticrainforest:
This is my initial attempt at creating a unified databrowserAPI. I am open for discussions.


pyproject.toml
Karinon, line 1: Add `rich` to the dependencies. It was not installed on my computer

Karinon (changes requested):
I have issues with the flavour:

```
localhost:7777/metadata_search/freva/file?flavour=cmip6
```

returns 

```
  File "/home/afast/miniconda3/envs/python311/lib/python3.11/site-packages/fastapi/routing.py", line 163, in run_endpoint_function
    return await dependant.call(**values)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/afast/workspace/FREVA/github-freva/databrowserAPI/src/databrowser/run.py", line 140, in metadata_search
    solr_search = SolrSearch(
                  ^^^^^^^^^^^
TypeError: databrowser.core.SolrSearch() got multiple values for keyword argument 'flavour'
```

Same happens for `uniq_keq`

antarcticrainforest:
> I have issues with the flavour:
> 
> ```
> localhost:7777/metadata_search/freva/file?flavour=cmip6
> ```
> 
> returns
> 
> ```
>   File "/home/afast/miniconda3/envs/python311/lib/python3.11/site-packages/fastapi/routing.py", line 163, in run_endpoint_function
>     return await dependant.call(**values)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/afast/workspace/FREVA/github-freva/databrowserAPI/src/databrowser/run.py", line 140, in metadata_search
>     solr_search = SolrSearch(
>                   ^^^^^^^^^^^
> TypeError: databrowser.core.SolrSearch() got multiple values for keyword argument 'flavour'
> ```
> 
> Same happens for `uniq_keq`

Good catch, those are path variables. If you give them as query parameters it'll fail. We can throw a meaningful error or silently try to pop the keys from the query params. Would a meaningful error message be better to teach people how to use it?

Karinon:
As already mentioned, I actually don't like the "ignore keys"-way. Meaningful error messages would be nice.

Karinon:
Another thing: I assume, that I will use the flavor-Thing always with `translate=false` on. I just tried it out on nextgems. It works on the facets (e.g. "ensembles" stay "ensembles" like in freva). But you still give me the translated primary facets. I would strongly prefer that the primary facets will not be translated as well. Otherwise you will force me to do some very ugly JSON-transformations.

Only thing I expect to change with `translate=false` ist the `facet_mapping`-Object

Karinon:
Two more things:

- `variables` is not a primary facet, I think it should be
- I would like to have the possibility to reduce the shown facets. The plugin-setup typically just queries a single facet like 'show me everything from facet `project` where the following search criteria are true'. This is not possible with the current API. Of course I could just filter it client-side but I don't want to as it wasted resources for nothing.

Karinon (approved):
LGTM

PR #4, opened 2024-01-29
Fixes issues with the intake API

Karinon:
Intake API created issues when the result had more than 150 results (fields were written as List of strings when they should be just String). This aims to fix this

### DIFF
--- a/src/databrowser/core.py
+++ b/src/databrowser/core.py
@@ -628,7 +628,9 @@ class SolrSearch:
             _, results = res
             for out in results.get("response", {}).get("docs", [{}]):
                 source = {
-                    k: out[k]
+                    k: out[k][0]
+                    if isinstance(out.get(k), list) and len(out.get(k)) == 1
+                    else out.get(k)
                     for k in [self.uniq_key] + self.translator.facet_hierachy
                     if out.get(k)
                 }


src/databrowser/core.py
```diff
@@ -628,7 +628,9 @@ async def _iterintake(self) -> AsyncIterator[str]:
             _, results = res
             for out in results.get("response", {}).get("docs", [{}]):
                 source = {
-                    k: out[k]
+                    k: out[k][0]
+                    if isinstance(out.get(k), list) and len(out.get(k)) == 1
+                    else out.get(k)
```
antarcticrainforest, line 633: Ok, now I get it. But is the og problem here or is this just a fix to another problem? As in, why is there a list if it's supposed to be a string? Did something go wrong when crawling metadata? A solr config issue?

Anyhoo I think this is fair enough!

antarcticrainforest (approved):
LGTM!

But my guess is there might be something going on deeper that is worth investigating.

src/databrowser/core.py
Karinon: I think the reason why (almost) everything is a list is because of the way how our Solr schema looks like, as almost everything is a single-element-list in there. I guess I am the one to blame for this decision as I had no idea how the data we ingest look like and just decided to make everything a list in the schema "just-in-case" so that nothing crashes.

It worked for the first ~150 Elements because you did the exact same thing I am doing in this PR for them in the initial Solr-call.

antarcticrainforest:
@Karinon the build process did not complete : Can you check what's going on? https://github.com/FREVA-CLINT/databrowserAPI/actions/runs/7694596668/job/20965688706

Karinon:
@antarcticrainforest The issue came with python 3.12 and I am sure, that the previous build does not work anymore aswell.

The error message is

```
  frozenlist/_frozenlist.c:8088:45: error: ‘PyLongObject’ {aka ‘struct _longobject’} has no member named ‘ob_digit’
   8088 |     const digit* digits = ((PyLongObject*)b)->ob_digit;
        |                                             ^~
  error: command '/usr/bin/gcc' failed with exit code 1


  at /usr/local/lib/python3.12/site-packages/poetry/installation/chef.py:164 in _prepare
      160│
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│
      163│             if error is not None:
    → 164│                 raise error from None
      165│
      166│             return path
      167│
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with frozenlist (1.3.3) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "frozenlist (==1.3.3)"'.
```

As of now I have no idea how to fix that except for setting the docker-container to `python:3.11`.

Is there any reason why you have set `poetry install -q --only main --no-root --no-directory --no-interaction` to quiet? It took me way too long to realize, that `-q` is the thing which eats the error message and it was quite frustrating. Can we get rid of it?

Btw: I still think that it makes sense to use actual version numbers instead of 'latest' if we really want to have a stable system :P

antarcticrainforest:
I'd say we get rid of poetry in favour for flit altogether. For my taste way too much boilerplate. Flit is simple and does the job how we need it. The project has been started by a former colleague of mine. I have already made a PR there as something didn't work as I wanted. Which was easy.

PR #50, opened 2024-07-24
Introduce exclusive search

antarcticrainforest:
As we discussed earlier. The restAPI is not able to perform a search the should NOT include certain values. This addresses this.

The old API realised an exclusive search by using `<facet_name>_not_=value`. I figured that it is more straightforward and allow for a more fine-grained search if we used normal lucene syntax. That is `facet_name=-value` or `facet_name=not value` or `facet_name=!value`. That way we don't have to deal with the facet checker.

If on the other hand, anyone says we should have the old way `<facet_name>_not_=value` please shout now.


freva-rest/src/databrowser_api/core.py
```diff
@@ -753,6 +764,28 @@ async def init_stream(self) -> Tuple[int, SearchResult]:
...
+        self, key: str, facets: List[str]
+    ) -> Tuple[str, str]:
+        """Create lucene search contain and NOT contain search queries"""
+
+        negative, positive = [], []
+        for search_value in facets:
+            if key not in self.uniq_keys:
+                search_value = search_value.lower()
+            if search_value.lower().startswith("not "):
+                search_value = "-" + search_value.lstrip("NOT ").lstrip("not ")
```
eelucio, line 777: should not be
```python
                negative.append(search_value.lstrip("NOT ").lstrip("not "))
```
instead ?

freva-rest/src/databrowser_api/core.py
```diff
@@ -753,6 +764,28 @@ async def init_stream(self) -> Tuple[int, SearchResult]:
...
 
+    def _join_facet_queries(
+        self, key: str, facets: List[str]
+    ) -> Tuple[str, str]:
+        """Create lucene search contain and NOT contain search queries"""
+
+        negative, positive = [], []
+        for search_value in facets:
+            if key not in self.uniq_keys:
+                search_value = search_value.lower()
```
eelucio, line 775: what happens with this case, is it lost? Could this be:
```python
                positive.append(search_value.lower())
```

instead?

freva-rest/src/databrowser_api/core.py
```diff
@@ -762,13 +795,11 @@ def _get_url(self) -> tuple[str, Dict[str, Any]]:
...
-            else:
-                search_value = " OR ".join(map(str.lower, value))
-            for char in self.escape_chars:
-                search_value = search_value.replace(char, "\\" + char)
-            query.append(f"{key.lower()}:({search_value})")
+            query_pos, query_neg = self._join_facet_queries(key, value)
+            if query_pos:
+                query.append(f"{key.lower()}:({query_pos})")
+            if query_neg:
+                query.append(f"-{key.lower()}:({query_neg})")
```
eelucio, line 802: I do not get it, if the negation is done in https://github.com/FREVA-CLINT/freva-nextgen/blob/668fcc086d72322d3047254819266b18f43d2025/freva-rest/src/databrowser_api/core.py#L777
what happens if you do it again here
```python
            if query_neg:
                query.append(f"-{key.lower()}:({query_neg})")
```

I added to query_neg so it would strip either not, NOT, ! or - 
and then add that minus to the query itself

docs/source/databrowser/cli.rst
```diff
@@ -108,10 +108,10 @@ There are many more options for defining a value for a given key:
 |                                                 | attribute1 AND value2  |
 |                                                 | for attribute2         |
 +-------------------------------------------------+------------------------+
-| ``attribute_not_=value``                        | Search for files NOT   |
+| ``attribute=-value`` ``attribute=not value``    | Search for files NOT   |
 |                                                 | containing value       |
 +-------------------------------------------------+------------------------+
-| ``attribute_not_=value1 attribute_not_=value2`` | Search for files       |
+| ``attribute=-value1 attribute=not value2``      | Search for files       |
```
eelucio, line 114: sorry for this dumb question, but has been the cli (freva core) changed so that is reflected? or does it go independently?

eelucio (changes requested):
I may be wrong but I think there is one small issue there.

on the other hand I do not think the syntax change in the cli will be breaking much (I would pressume that this is a fairly unused capability)

docs/source/databrowser/cli.rst
```diff
@@ -108,10 +108,10 @@ There are many more options for defining a value for a given key:
 |                                                 | attribute1 AND value2  |
 |                                                 | for attribute2         |
 +-------------------------------------------------+------------------------+
-| ``attribute_not_=value``                        | Search for files NOT   |
```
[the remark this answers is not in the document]
MoSHad91: and yes, it seems it covers curly braces for neg as well: https://github.com/FREVA-CLINT/freva-nextgen/blob/contains-not/freva-rest/src/databrowser_api/core.py#L786

MoSHad91 (approved):
this is one just a comment on the already developed part which has not been documented ... lgtm🚀

freva-rest/src/databrowser_api/core.py
antarcticrainforest: We don't need that. because `search_value` is already transformed to lowercase when the search key is not file or uri. We do that because files/uri stored without making the lowercase (we need to access them)

freva-rest/src/databrowser_api/core.py
antarcticrainforest: Yes, I know the logic is kind of awkward. The idea is to replace leading *not*'s by leading -. I guess I could aswell do it like this:

```python
if search_value.lower().startswith("not "):
    negative.append(search_value.lstrip("NOT ").lstrip("not "))
elif search_value[0] in ("!", "-"):
     negative.append(search_value.lstrip("!").lstrip("-"))
else:
     positive.append(search_value)

```

freva-rest/src/databrowser_api/core.py
antarcticrainforest: We are not doing it again. My idea was to collect negative search statements (searches that should NOT contain the search result) and positive ones (facets that should contain search results). Once we have those we need to translate this to lucene syntax. Unfortunately, this doesn't work in lucene:

`{"fq": "project":"-cmip6"} ` it needs to be `{"fq":"-project":"cmip6"}`.

I am not super happy about this but there is nothing we can do.

In my opinion, something like project=-cmip6 is easy to understand and it is easier to implement than putting the not condition to the search key, because this: -project=cmip6 doesn't work in python nor at the cli. I guess this is why the old system had the` _not_ `construct. project_not_ = cmip.


We can reintroduce this for the sake of backward compatibility. But then we'd have to change the facet checker. Might not be a big deal though? Any opnios @MoSHad91 @eelucio

docs/source/databrowser/cli.rst
antarcticrainforest: That's a very good question. No, it hasn't yet. I think we should switch to using the databrowser api also in the old freva core.

antarcticrainforest:
I've also added the `facet_not_=value` construct back to the code. So there is nothing to be done from the plugin side.

The side effect it that it doesn't matter where the `_not_` is located so `_not_facet=value` or `facet_not_=value` will have the same effect even `fa_not_cet=value` if you think that is problematic let me know.

eelucio:
LGTM!

One last dumb minor question:

Is there any reason why you define a function [`foreward_lookup()`](https://github.com/FREVA-CLINT/freva-nextgen/blob/contains-not/freva-rest/src/databrowser_api/core.py#L263) instead of e.g. `forward_lookup()`

If I want to do some minor typo correction should I open a new branch once this is merge, or wait untill all is merge?

eelucio:
> I've also added the `facet_not_=value` construct back to the code. So there is nothing to be done from the plugin side.
> 
> The side effect is, that it doesn't matter where the `_not_` is located so `_not_facet=value` or `facet_not_=value` will have the same effect even `fa_not_cet=value` if you think that is problematic let me know.

The fact that we have some facets written with underscore e.g. [`file_no_version` in the managed schema](https://github.com/FREVA-CLINT/freva-service-config/blob/main/solr/managed-schema.xml#L65) should not be a problem, right?

I guess that for the future we will just need to be sure that nobody enters a facet that has a `_not_` in between the name
as the above mentioned param is dangerously close to this statement. But as far as I know it was not searchable, right?


otherwise LGTM!

one (very) minor comment:
is there any reason why there is a function called [`foreward_lookup()`](https://github.com/FREVA-CLINT/freva-nextgen/blob/contains-not/freva-rest/src/databrowser_api/core.py#L263) (old english) instead of `forward_lookup()`? I hope the typo correctors do not change it automatically 

and one last one, shall I'd like to correct some typos (e.g. in docstrings etc.) should I wait until many other branches are merged?

PR #54, opened 2024-07-30
Open id connect

antarcticrainforest:
- These changes untie the relation of the auth mechanism to keycloak
  rather than using exclusively keycloak to API should be able to work
  with any kind of open connect id provider.
- The client uses authlib for authentication.

### DIFF
--- a/docs/source/auth/index.rst
+++ b/docs/source/auth/index.rst
@@ -77,11 +77,10 @@ included in the authorization header for secured endpoints.
         {
             "access_token": "eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6.."
             "token_type": "Bearer",
-            "expires_in": 300,
+            "expires": 1722874908,
             "refresh_token": "eyJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia2lkIi.."
-            "refresh_expires_in": 1800,
-            "scope": "email",
-            "not-before-policy": 0
+            "refresh_expires": 1722876408,
+            "scope": "profile email address",
         }
 
     Code examples
--- a/docs/source/auth/index.rst
+++ b/docs/source/auth/index.rst
@@ -37,11 +37,17 @@ included in the authorization header for secured endpoints.
     :type username: str
     :form password: The password for the login
     :type password: str
-    :form client-id: The unique identifier for your application used to
+    :form refresh_token: The refresh token that is used to create a new token
+                         the refresh token can be used instead of authorizing
+                         via user creentials.
+    :type refresh_token: str
+    :form client_id: The unique identifier for your application used to
                      request an OAuth2 access token from the authentication
                      server, this form parameter is optional.
-    :type client-id: str
-
+    :type client_id: str
+    :form client_secret: An optional client secret used for authentication
+                         this parameters is optional an in most cases not neeede
+    :type client_secret: str
     :statuscode 200: no error
     :statuscode 401: unauthorized
     :resheader Content-Type: ``application/json``: access and refresh token.
@@ -73,7 +79,9 @@ included in the authorization header for secured endpoints.
             "token_type": "Bearer",
             "expires_in": 300,
             "refresh_token": "eyJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia2lkIi.."
-            "refresh_expires_in": 1800
+            "refresh_expires_in": 1800,
+            "scope": "email",
+            "not-before-policy": 0
         }
 
     Code examples
@@ -144,129 +152,6 @@ included in the authorization header for secured endpoints.
                     curl_easy_setopt(curl, CURLOPT_URL, "https://www.freva.dkrz.de/api/auth/v2/token");
                     curl_easy_setopt(curl, CURLOPT_HTTPHEADER, headers);
                     curl_easy_setopt(curl, CURLOPT_POSTFIELDS, "username=janedoe&password=janedoe123");
-
-                    res = curl_easy_perform(curl);
-                    curl_easy_cleanup(curl);
-                }
-                curl_global_cleanup();
-                return 0;
-            }
-
----
-
-
-.. http:post:: /api/auth/v2/refresh
-
-    Renew the access token with help of a refresh token.
-
-    :form refresh-token: The refresh token used to refresh the access token
-    :type refresh-token: str
-    :form client-id: The unique identifier for your application used to
-                     request an OAuth2 access token from the authentication
-                     server. This form parameter is optional
-    :type client-id: str
-
-    :statuscode 200: no error
-    :statuscode 401: unauthorized
-    :resheader Content-Type: ``application/json``: access and refresh token.
-
-
-    Example Request
-    ~~~~~~~~~~~~~~~
-
-    .. sourcecode:: http
-
-        POST /api/auth/v2/refresh HTTP/1.1
-        host: www.freva.dkrz.de
-        Content-Type: application/json
-
-        {
-            "refresh_token": "your_refresh_token"
-        }
-
-
-    Example Request
-    ~~~~~~~~~~~~~~~
-
-    .. sourcecode:: http
-
-        HTTP/1.1 200 OK
-        Content-Type: application/json
-
-        {
-            "access_token": "eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6.."
-            "token_type": "Bearer",
-            "expires_in": 300,
-            "refresh_token": "eyJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia2lkIi.."
-            "refresh_expires_in": 1800
-        }
-
-    Code examples
-    ~~~~~~~~~~~~~
-    Below you can find example usages of this request in different scripting and
-    programming languages
-
-    .. tabs::
-
-        .. code-tab:: bash
-            :caption: Shell
-
-            curl -X POST https://www.freva.dkrz.de/api/auth/v2/refresh \
-             -d "refresh-token=eyJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia..."
-
-        .. code-tab:: python
-            :caption: Python
-
-            import requests
-            response = requests.post(
-              "https://www.freva.dkrz.de/api/auth/v2/refresh",
-              data={"refresh-token": "yJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia..."}
-            )
-            token_data = response.json()
-
-        .. code-tab:: r
-            :caption: gnuR
-
-            library(httr)
-
-            response <- POST(
-               "https://www.freva.dkrz.de/api/auth/v2/refresh",
-               body = list(refresh-token = "yJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia."),
-               encode = "form"
-            )
-            token_data <- content(response, "parsed")
-
-        .. code-tab:: julia
-            :caption: Julia
-
-            using HTTP
-            using JSON
-
-            response = HTTP.POST(
-              "https://www.freva.dkrz.de/api/auth/v2/refresh",
-              body = Dict("refresh-token" => "yJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia.")
-            )
-            token_data = JSON.parse(String(response.body))
-
-        .. code-tab:: c
-            :caption: C/C++
-
-            #include <stdio.h>
-            #include <curl/curl.h>
-
-            int main() {
-                CURL *curl;
-                CURLcode res;
-
-                curl_global_init(CURL_GLOBAL_DEFAULT);
-                curl = curl_easy_init();
-                if(curl) {
-                    struct curl_slist *headers = NULL;
-                    headers = curl_slist_append(headers, "Content-Type: application/x-www-form-urlencoded");
-
-                    curl_easy_setopt(curl, CURLOPT_URL, "https://www.freva.dkrz.de/api/auth/v2/refresh");
-                    curl_easy_setopt(curl, CURLOPT_HTTPHEADER, headers);
-                    curl_easy_setopt(curl, CURLOPT_POSTFIELDS, "refresh-token=yJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia...");
 
                     res = curl_easy_perform(curl);
                     curl_easy_cleanup(curl);
--- a/docs/source/auth/index.rst
+++ b/docs/source/auth/index.rst
@@ -45,8 +45,8 @@ included in the authorization header for secured endpoints.
                      request an OAuth2 access token from the authentication
                      server, this form parameter is optional.
     :type client_id: str
-    :form client_secret: An optional client secret used for authentication
-                         this parameters is optional an in most cases not neeede
+    :form client_secret: An optional client secret used for authentication.
+                         This param. is optional and in most cases not needed
     :type client_secret: str
     :statuscode 200: no error
     :statuscode 401: unauthorized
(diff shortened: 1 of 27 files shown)


MoSHad91:
i u don't mind, on the weekend I will have a look, I'm busy with developing things for regiklim at the moment

antarcticrainforest:
Hmm, I actually need to deploy this on a machine for beta testing. For this I need to change the deployment, ideally, I'd like to work with a version that isn't outdated.

antarcticrainforest:
I am going to push a dev version to the container registry and use this.

freva-rest/src/freva_rest/__init__.py
```diff
@@ -1,7 +1,7 @@
 import os
 from pathlib import Path
 
-__version__ = "2403.0.3"
+__version__ = "2408.0.0-dev2"
```
MoSHad91, line 4: there is no standard way to make it dynamic to inherits from the `freva-rest` version [here](https://github.com/FREVA-CLINT/freva-nextgen/blob/open-id-connect/freva-client/src/freva_client/__init__.py#L20) or vice versa? generally I'm not fan of defining version in `__init__.py` and instead of defining version in diff files, my suggestion is to define it in `pyproject.toml` and build the image based on the version in pyproject.

freva-rest/src/freva_rest/auth.py
```diff
@@ -1,175 +1,93 @@
...
-    """Get the status of an existing user token."""
-    return await get_current_user(token, verify_exp=False)
-
-
-@app.post(
-    "/api/auth/v2/refresh", response_model=Token, tags=["Authentication"]
-)
-async def refresh_access_token(
+@app.post("/api/auth/v2/token", tags=["Authentication"])
+async def fetch_or_refresh_token(
```
MoSHad91, line 60: instead of parsing the request before fetch_or_refresh_token stage, if we get the whole request as Request here, we don't need to define and take care of so many attrs complicated typing i think:

```python
from fastapi import Request
async def fetch_or_refresh_token(request: Request):
    data = {
        "client_id": request.client_id or server_config.oidc_client,
        "client_secret": request.client_secret or os.getenv("OIDC_CLIENT_SECRET", ""),
        "grant_type": request.grant_type,
        "username": request.username,
        "password": request.password,
        "refresh_token": request.refresh_token,
    }
    data = {k: v for k, v in data.items() if v}
```

tbh I haven't tried in fastapi but it seems it has to be feasible ...

freva-rest/src/freva_rest/auth.py
```diff
@@ -178,35 +96,35 @@ async def refresh_access_token(
...
+        "client_id": (client_id or "").replace("None", "") or server_config.oidc_client,
+        "client_secret": client_secret or os.getenv("OIDC_CLIENT_SECRET", ""),
+        "grant_type": grant_type,
+    }
+    if grant_type == "password":
+        data["password"] = password
+        data["username"] = username
+    else:
+        data["refresh_token"] = refresh_token
+    async with aiohttp.ClientSession(timeout=TIMEOUT) as client:
```
MoSHad91, line 119: another point, in data-loader we perfectly consider ssl certs to secure the connection but here we don't use it here. But the problem which is gonna happen, i think it would be in `dev`, which we need to consider an env for this:


```python
ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
ssl_context.load_cert_chain('/path/to/cert.pem', keyfile='/path/to/key.pem')
async with aiohttp.ClientSession(timeout=TIMEOUT, connector=aiohttp.TCPConnector(ssl=ssl_context)) as client:
```
which `/path/to/cert.pem` and `/path/to/key.pem` can comes from env. because we use them in main function of cli of dataportal as well ...

https://medium.com/@mariovanrooij/adding-https-to-fastapi-ad5e0f9e084e

freva-rest/src/freva_rest/cli.py
```diff
@@ -154,9 +159,10 @@ def start(
             asyncio.run(read_data(core, cfg.solr_host, cfg.solr_port))
     workers = {False: int(os.environ.get("API_WORKER", 8)), True: None}
     ssl_cert, ssl_key = get_cert_file(ssl_cert_dir, ssl_cert, ssl_key)
-    keycloak_client_id = os.getenv("KEYCLOAK_CLIENT_ID", "freva")
-    keycloak_client_secret = os.getenv("KEYCLOAK_CLIENT_SECRET", "")
+    oidc_client_id = os.getenv("OIDC_CLIENT_ID", "freva")
+    oidc_client_secret = os.getenv("OIDC_CLIENT_SECRET", "")
     api_services = ",".join(services).replace("_", "-")
+    oidc_url = "http://localhost:8080/realms/freva/.well-known/openid-configuration"
```
MoSHad91, line 165: beautiful  👍

MoSHad91 (commented):
Excelent @antarcticrainforest 

just dropped some minor comments, but after clarification of Token typedict it looks good to me

freva-client/src/freva_client/auth.py
```diff
@@ -78,26 +70,20 @@ def _refresh(
         self, url: str, refresh_token: str, username: Optional[str] = None
     ) -> Token:
         """Refresh the access_token with a refresh token."""
+        auth = self._auth_cls.refresh_token(f"{url}/token", refresh_token or " ")
         try:
-            res = requests.post(
-                f"{url}/refresh",
-                data={"refresh-token": refresh_token},
-                timeout=5,
+            return self.set_token(
```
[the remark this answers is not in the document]
MoSHad91: ahhh I got where the problem comes form! In both `freva_rest` and `freva_client` you defined Token with different structure and attrs which I have no clue why we don't define the centralized Token class from the originated location and then use in different endpoints. It made me a bit confused in the first tbh

docs/source/auth/index.rst
```diff
@@ -37,11 +37,17 @@ included in the authorization header for secured endpoints.
...
+                         via user creentials.
+    :type refresh_token: str
+    :form client_id: The unique identifier for your application used to
                      request an OAuth2 access token from the authentication
                      server, this form parameter is optional.
-    :type client-id: str
-
+    :type client_id: str
+    :form client_secret: An optional client secret used for authentication
+                         this parameters is optional an in most cases not neeede
```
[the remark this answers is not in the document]
antarcticrainforest: Ok

docs/source/auth/index.rst
```diff
@@ -73,7 +79,9 @@ included in the authorization header for secured endpoints.
             "token_type": "Bearer",
             "expires_in": 300,
             "refresh_token": "eyJhbGciOiJIUzUxMiIsInR5cCIgOiAiSldUIiwia2lkIi.."
-            "refresh_expires_in": 1800
+            "refresh_expires_in": 1800,
+            "scope": "email",
+            "not-before-policy": 0
```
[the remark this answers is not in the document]
antarcticrainforest: This is the response from keycloak, therefore I added a response as they come from keycloak.

freva-client/src/freva_client/auth.py
```diff
@@ -32,31 +32,23 @@ def __new__(cls) -> "Auth":
...
-            return 0.0
-
-    def _token_is_expired(self, url: str, token: str) -> bool:
-        """Check if the token has already expired."""
-        exp = datetime.fromtimestamp(self.token_expiration_time(url, token))
-        return datetime.now() > exp
+        if self._auth_token is None:
+            exp = 0.0
+        else:
+            exp = self._auth_token["expires"]
```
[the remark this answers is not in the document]
antarcticrainforest: Is that the case? TBH I would be more pragmatic here and only do something about this when a problem occurs.

freva-client/src/freva_client/auth.py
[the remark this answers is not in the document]
antarcticrainforest: The freva rest code defines the token dat as it comes from the keycloak - I didnt' want to change it. The client lib makes an absolute expiry timestamp from the relative expires in value. We could let the restAPI return a json payload that already defines the absolute token expiry time. But I would like to keep the interference of the restAPI with the auth system as little as possible.

freva-client/src/freva_client/auth.py
```diff
@@ -118,27 +104,20 @@ def check_authentication(self, auth_url: Optional[str] = None) -> Token:
...
-                f"{auth_url}/token",
-                data={
-                    "username": username,
-                    "password": getpass(pw_msg),
-                },
-                timeout=5,
+            return self.set_token(
+                access_token=auth["access_token"],
+                token_type=auth["token_type"],
+                expires_in=auth["expires_in"],
```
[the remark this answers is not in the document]
antarcticrainforest: See my comment above.

freva-rest/src/freva_rest/__init__.py
antarcticrainforest: I would like to be able to do this:

```python
import foo
print(foo.__version__)
```

This is important, because when people report issues the first we want to ask is what version they are using. 

I know what you could muck around with pkg_resources or importlib. But I think it's very clunky. Hence I will keep this __version__ constructs.

freva-rest/src/freva_rest/auth.py
antarcticrainforest: Yes, that was indeed my first approach. But then I decided to follow this approach because it integrates much better with the swagger help.

freva-rest/src/freva_rest/auth.py
antarcticrainforest: I am not sure I understand this. The data-loader isn't involved here.

The ssl connection to the keycloak shouldn't be based on a self-signed cert. 

On the other hand for the redis connection in the data-loader, we want to use a self-signed cert to make sure that only hosts that hold the private key can make connections to the redis host.

antarcticrainforest (commented):
Ok thanks for looking at this.

The Question about the Token. I think the `expires_in` data that comes from keycloak is pretty useless as we have to know when the token was issued. But better would be a absolute `expires_at` value. We could already convert the `expires_in` to a `expiries_at` on the restAPI side. But I wanted to keep the interference of the restAPI with the auth service as minimal as possible - that is why I decided to simply pass the token data as it was recieved by the auth instance.

MoSHad91 (approved):

MoSHad91:
@antarcticrainforest 

I was in the middle of reviewing `freva-config` and didn't know where to write this. So i write here which seems more relevant to this:

Alright, if you would like to keep it as-is, I think `freva-config` repo has to be merged prior to `freva-nextgen` and beforehand some changes in this repo is needed. it seems you moved `keyclok` dir in freva-nextgen to `freva-config` and we don't need to have same files in both `freva-nextgen/dev-env/keycloak` and `freva-nextgen/dev-env/config/keycloak`. It needs some changes in https://github.com/FREVA-CLINT/freva-nextgen/blob/open-id-connect/dev-env/docker-compose.yaml as well
Same story for `redis-cmd.sh`.

But in principal, if keycloak is just designed for dev configuration, it is completely non-relevant to the definition/description of `freva-nextgen` repo.  Or you can at least put the whole `dev-utils.py` `keycloak` dev dir and so on in a same dev dir. 

> This repository holds all definitions for Docker images to create services that are needed to run Freva in production and development mode

also i think you have to move the https://github.com/FREVA-CLINT/freva-nextgen/blob/open-id-connect/dev-env/docker-compose.yaml to `freva-config` dev dir and also remove the `dev-env/key.py` from here as you made `dev-utils.py` in `freva-config`

antarcticrainforest:
> @antarcticrainforest
> 
> I was in the middle of reviewing `freva-config` and didn't know where to write this. So i write here which seems more relevant to this:
> 
> Alright, if you would like to keep it as-is, I think `freva-config` repo has to be merged prior to `freva-nextgen` and beforehand some changes in this repo is needed. it seems you moved `keyclok` dir in freva-nextgen to `freva-config` and we don't need to have same files in both `freva-nextgen/dev-env/keycloak` and `freva-nextgen/dev-env/config/keycloak`. It needs some changes in https://github.com/FREVA-CLINT/freva-nextgen/blob/open-id-connect/dev-env/docker-compose.yaml as well Same story for `redis-cmd.sh`.
> 
> But in principal, if keycloak is just designed for dev configuration, it is completely non-relevant to the definition/description of `freva-nextgen` repo. Or you can at least put the whole `dev-utils.py` `keycloak` dev dir and so on in a same dev dir.
> 
> > This repository holds all definitions for Docker images to create services that are needed to run Freva in production and development mode
> 
> also i think you have to move the https://github.com/FREVA-CLINT/freva-nextgen/blob/open-id-connect/dev-env/docker-compose.yaml to `freva-config` dev dir and also remove the `dev-env/key.py` from here as you made `dev-utils.py` in `freva-config`

Hmm,

yes I will need to change the freva-nexgen, but obviously after this here has been merged.

it's not only used for production but also for development. This is why we used this repo as a submodule. In freva-web and freva. 

If you don't like the approach that keycloak and open ldap is on the same level as the other services I can move it to a dev folder to make in more clear that's dev. But imho I don't think that it harms if this  definitions aboard even if they are not used in production.

About the docker-compose files. I thought about this but ditched the idea because if you wanted to make the smallest change, like adding a new env var. You'd have to submit a PR here and wait for it to be merged.  This is too painful if you ask me. So I decided to leave the docker-compose files where they are. But I'll think about something.

antarcticrainforest:
I've unified the token responses in the rest api so that it matches to token in the client. It also works with authlib.

freva-client/src/freva_client/auth.py
```diff
@@ -36,33 +37,35 @@ def __init__(self) -> None:
...
         access_token: str,
         refresh_token: Optional[str] = None,
         expires_in: int = 10,
         refresh_expires_in: int = 10,
-        expires: Optional[float] = None,
-        refresh_expires: Optional[float] = None,
+        expires: Optional[Union[float, int]] = None,
+        refresh_expires: Optional[Union[float, int]] = None,
         token_type: str = "Bearer",
+        scope: str = "profile email address",
```
MoSHad91, line 57: https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims:~:text=scope%3Dopenid%20profile%20email%20phone

freva-rest/src/freva_rest/auth.py
```diff
@@ -127,4 +127,26 @@ async def fetch_or_refresh_token(
             logger.error(error)
             raise HTTPException(status_code=404)
         token_data = await response.json()
-    return Token(**token_data)
+        expires_at = (
```
MoSHad91, line 130: hmm, I'm not against of defining so many ORs for different `expires` keywords, but if you really would like to make it scaleable for the future, go for the approach that I mentioned in first review and defining a expiration_keys dictionary. 

https://github.com/FREVA-CLINT/freva-nextgen/pull/54#discussion_r1703859375

MoSHad91 (approved):
lgtm🚀

freva-rest/src/freva_rest/auth.py
antarcticrainforest: That's not really true. The piece of code you posted, a) loops every time through keys and then b) checks for each entry if the key is within the in hash map (aka dict) that's. Under the hood that's another loop. And hence O(n^2). Retrieving items from a hash map is O(1) checking their presence O(n). So it's always better to access from a hashmap rather than checking the presence in a hashmap. Take a look at this:


```python
 def foo_get():
    lookup = {"expires_in": 200}
    return lookup.get("exp") or lookup.get("expires") or lookup.get("expires_at") or lookup.get("expires_in")

def foo_loop():
   lookup = {"expires_in": 200}
   keys = ["exp", "expires", "expires_at", "expires_in"]
   for key in keys:
      if key in lookup:
         return lookup[key]
         
%timeit foo_get()
460 ns ± 53.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

%timeit foo_loop()
726 ns ± 128 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

PR #59, opened 2024-08-07
Use new freva-service config.

antarcticrainforest:
This should be straight forward. It only reflects the changes from the PR in freva-service-config.


MoSHad91 (approved):
As we all discussed changes in the last PR, when you got green light from CI then LGTM 🚀

antarcticrainforest:
Sorry, I will need to add more changes to the freva-sevice repo!

MoSHad91:
i think u need to update the submodule here to the latest hash ...

antarcticrainforest:
I hope that should do it.

PR #64, opened 2024-08-14
Fix cloudpickle load bug.

antarcticrainforest:
I enabled the data-loader in freva-dev  an noticed that the loading failed the code rest-api (not the data-loader) wanted to import xarray and all the backends. This was because I added the cloudpickle dataset object to the cache. 

This cached dataset is used internally by the data-loader code, not by the rest-api. Simple solution: add the xarray dataset to a dedicated cache entry. The problem should be solved.

What bothers me a little bit is that the tests couldn't fetch this as the code is running in the same python environment and hence all libraries are available. But I guess we'll have to live with that.


freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -4,20 +4,16 @@
...
 from dask.distributed import LocalCluster  # fades distributed
 from dask.distributed import Client
 from xarray.backends.zarr import encode_zarr_variable
-from xpublish.utils.zarr import (  # fades xpublish
-    create_zmetadata,
-    encode_chunk,
-    get_data_chunk,
-    jsonify_zmetadata,
-)
+from xpublish.utils.zarr import create_zmetadata  # fades xpublish
```
MoSHad91, line 15: would be more clear to turn the changes back here

freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -185,16 +176,24 @@ def from_object_path(
         ).dict()
         expires_in = str_to_int(os.environ.get("API_CACHE_EXP"), 3600)
         status_dict["status"] = 3
+        dset: Optional[xr.Dataset] = None
         self.cache.setex(path_id, expires_in, cloudpickle.dumps(status_dict))
```
MoSHad91, line 180: changing the name here as `self.cache.setex(f"{path_id}-metadata", expires_in, cloudpickle.dumps(status_dict))` could be an idea to not get confused in bug fixing.

freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -185,16 +176,24 @@ def from_object_path(
...
-            status_dict["obj"] = dset
             status_dict["meta"] = metadata
             status_dict["status"] = 0
+            # We need to add the xarray to an extra cache entry because the
+            # status_dict will be loaded by the rest-api, if the xarray dataset
+            # is present the rest-api code will attempt to instanciate the
+            # pickled dataset object and that might fail because we might or
+            # might not have xarray and all the backends instanciated.
+            # Since the xarray dataset object isn't needed anyway byt the
+            # rest-api we simply add it to a cache entry of its own.
```
MoSHad91, line 195: as status dict is changing here, what do u think to update srtting the metadat var in cache before `{path_id}-dset`? i mean adding this line `self.cache.setex(f"{path_id}-metadata", expires_in, cloudpickle.dumps(status_dict))` before dumping the dset

freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -247,9 +243,12 @@ def load_object(self, key: str) -> LoadDict:
         KeyError: If the key doesn't exist in the cache (anymore).
         """
         data_cache = self.cache.get(key)
-        if data_cache is None:
+        dset_bin = self.cache.get(f"{key}-dset")
```
MoSHad91, line 246: and to make the error catching smarter, we can handle it via checking the metadata var in cache as well. i meant, if it has metadata and not dset, raise another error ... just popped up

freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -247,9 +243,12 @@ def load_object(self, key: str) -> LoadDict:
         KeyError: If the key doesn't exist in the cache (anymore).
         """
         data_cache = self.cache.get(key)
```
MoSHad91, line 245: i thinnk we have to keep a better naming convention here to make it more understandable for first time developer who jump to fix the bug. like, `dset_cache` and `meta_cache`

freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -247,9 +243,12 @@ def load_object(self, key: str) -> LoadDict:
...
-        if data_cache is None:
+        dset_bin = self.cache.get(f"{key}-dset")
+        if data_cache is None or dset_bin is None:
             raise KeyError(f"{key} uuid does not exist (anymore).")
-        return cast(LoadDict, cloudpickle.loads(data_cache))
+        load_dict = cast(LoadDict, cloudpickle.loads(data_cache))
```
MoSHad91, line 249: same naming convention here as well

MoSHad91 (approved):
generally the approach of handling dset lgtm, just some naming stuff has to be more considered

freva-data-portal-worker/src/data_portal_worker/load_data.py
antarcticrainforest: That wasn't me, that was my linter.

freva-data-portal-worker/src/data_portal_worker/load_data.py
antarcticrainforest: Unfortunately, that's not as simple. Because this also needs to be handled on the rest-api side. I'd like to keep this as this this for the time being.

freva-data-portal-worker/src/data_portal_worker/load_data.py
antarcticrainforest: No, we shouldn't do that here. Because if something goes wrong we need to update the status dict that something goes wrong.


On the other hand I cannot add the xarray dataset after the try execpt block because I cannot add None  to the cache. I could pickle the None and then add it. But then I have to get the dataset unpickle it and check if it's None. To me its just more straight forward to only add it if its good and then simply check for it's pure existence in the cache.

freva-data-portal-worker/src/data_portal_worker/load_data.py
antarcticrainforest: Ok

PR #69, opened 2024-08-28
Add no rich option

antarcticrainforest:
This fixes broken help text in the docs.


MoSHad91 (approved):
lgtm after passing the test ci

PR #87, opened 2024-10-17
Add setuptools

antarcticrainforest:
Something has changed upstream with pip. intake-esm needs setup tools, but it's no longer part of the python packaging. So we install it here. In metadata-inspector the theme thing happened: https://github.com/conda-forge/metadata-inspector-feedstock/pull/9

### DIFF
--- a/freva-client/pyproject.toml
+++ b/freva-client/pyproject.toml
@@ -28,6 +28,7 @@ dependencies = [
 "requests",
 "intake_esm",
 "rich",
+"setuptools",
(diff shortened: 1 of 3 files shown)


antarcticrainforest:
@mo-dkrz, it seems  to work again, opposed to previous attempt: https://github.com/FREVA-CLINT/freva-nextgen/actions/runs/11383863141/job/31670378746

mo-dkrz (approved):

PR #89, opened 2024-10-17
Add user data

### LINKED ISSUE #32: Create PUT methods for the databrowser API
# Improving the freva-rest API

In the existing freva system, users are able to add custom metadata to Apache Solr, which has been quite popular. To enhance this functionality in the new freva-rest API (databrowser API), we propose implementing a PUT method. This method will allow users to add their own metadata seamlessly. The new PUT endpoint will validate against the schema and integrate the metadata into Apache Solr.

## Avoiding File Name Creation

The current UserData class in freva manages metadata by merging input files with user-provided metadata, resulting in a new file structure under the crawl_my_data root directory. However, this approach necessitates creating new file names, often accomplished through soft links, copying, or moving data, which introduces complications. To streamline this process for the metadata-crawler, we propose extending the PUT method to also update MongoDB, linked with the REST API. This way, the metadata-crawler can access metadata directly from MongoDB, eliminating the need for new file names.

## Introducing a user Facet

Additionally, we recommend implementing a new user facet in Solr. This enhancement enables users to manage their project data independently without affecting the primary metadata. Currently, we use complex naming conventions like `project=<username>.` user project, which can be cumbersome. With the user facet, users can seamlessly index their project data under their own namespace within Solr. This change not only simplifies data management but also allows for easy indexing of intake-esm catalogue files. All catalogue entries can be grouped under theuser` facet, ensuring they are distinct from the main metadata collection and eliminating any risk of interference.

mo-dkrz:
This PR introduces two updates related to the `PUT` and `DELETE` methods for adding and removing user data.

For adding user data, the process involves three stages. First, it checks if the provided path exists. Then, it iterates through the path, crawling metadata for each data. Finally, the metadata is ingested into both Solr and Mongo using a multithreaded task. A [validator](https://github.com/FREVA-CLINT/freva-service-config/blob/add-mongo-validator/mongo/mongo-userdata-init.js) for Mongo is also in place to ensure data adheres to a pre-defined schema, consistent with Solr.

The delete is straightforward, retrieving the relevant facets and deleting them from both Solr and Mongo without multithreading.

Next steps include 1. developing a new Mongo backend for the metadata crawler and 2. updating the Mongo serrvice of Docker Compose freva-rest playbook in Freva deployment.

This PR also includes minor changes such as adding a job to check for pre-release tags and switching the Dockerfile to Miniconda for easier installation of netCDF and HDF libraries for harvesting data and metadata.

A final commit is left. 1. changing the branch in `dev-env/config` to main. 2. remove letter `v` from the created image in `build_job` ci
Closes #32


freva-client/src/freva_client/query.py
```diff
@@ -733,17 +712,111 @@ def _facet_search(
             constraints = data["primary_facets"]
         return {f: v for f, v in data["facets"].items() if f in constraints}
 
-    def _get(
-        self, url: str, **kwargs: Any
-    ) -> Optional[requests.models.Response]:
+    def add_user_data(
+        self, username: str, paths: List[str], facets: Dict[str, str]
+    ) -> None:
+        """Add user data to the databrowser.
```
antarcticrainforest, line 718: In the cli you used the term "into", "to" is also correct. But let us stick to one, though.

freva-client/src/freva_client/query.py
```diff
@@ -733,17 +712,111 @@ def _facet_search(
...
+
+        Via this functionality, user would be able to add data to the databrowser.
+        It accepts file paths and metadata facets to categorize and store the user's
+        data.
```
antarcticrainforest, line 722:
```suggestion
        This functionality allows users to add data into the databrowser.
        The method accepts file paths and metadata search attributes to categorize and store the user's
        data.
```

freva-client/src/freva_client/query.py
```diff
@@ -733,17 +712,111 @@ def _facet_search(
...
+                ["."],
+                {"project": "cmip5", "experiment": "something"}
+            )
+        """
+        url = f"{self._cfg.userdata_url}/{username}"
+        token = self._auth.check_authentication(auth_url=self._cfg.auth_url)
+        headers = {"Authorization": f"Bearer {token['access_token']}"}
+        params = {"paths": paths}
+        if "username" in facets:
+            del facets["username"]
```
antarcticrainforest, line 762:
```suggestion
        facets.pop("username", None)
```

freva-client/src/freva_client/query.py
```diff
@@ -733,17 +712,111 @@ def _facet_search(
...
+
+        if result is None:
+            raise ValueError("Failed to add user data")
+
+    def delete_user_data(self, username: str, search_keys: Dict[str, str]) -> None:
+        """
+        Delete user data from the databrowser.
+
+        Uing this, user would be able to delete the user's data from the databrowser
+        based on the provided search keys.
```
antarcticrainforest, line 774:
```suggestion
        Delete user data from the databrowser.

       This function allows users to delete their data from the databrowser
       based on the specified search keys.
```

freva-client/src/freva_client/query.py
```diff
@@ -733,17 +712,111 @@ def _facet_search(
...
+        ~~~~~~~~~~
+        username: str
+            The username associated with the data to be deleted.
+        search_keys: dict[str, str]
+            A dictionary containing the search keys to identify the data to be deleted.
+
+        Returns
+        ~~~~~~~~
+        None
+            If the operation is successful, no return value is provided.
```
antarcticrainforest, line 786: I think you can delete that.

freva-client/src/freva_client/query.py
```diff
@@ -733,17 +712,111 @@ def _facet_search(
...
+        ~~~~~~~
+        .. execute_code::
+
+            from freva_client import databrowser, authenticate
+            token_info = authenticate(username="janedoe")
+            db = databrowser()
+            db.delete_user_data(
+                "janedoe",
+                {"project": "cmip5", "experiment": "something"}
+            )
```
antarcticrainforest, line 803: I haven't look at it yet, but if the restAPI supports wild-cards you can also show it here.

freva-client/src/freva_client/query.py
```diff
@@ -757,3 +830,51 @@ def _get(
...
+            return res
+        except KeyboardInterrupt:
+            pprint("[red][b]User interrupt: Exit[/red][/b]", file=sys.stderr)
+
+        except (
+            requests.exceptions.ConnectionError,
+            requests.exceptions.HTTPError,
+        ) as error:
+            msg = f"adding user data request failed with {error}"
+            logger.warning(msg)
```
antarcticrainforest, line 860: I am not sure, how we want to deal with errors. In some instances, it might useful to throw an error, in some not - think about how you want to deal with that.

freva-client/src/freva_client/query.py
```diff
@@ -757,3 +830,51 @@ def _get(
...
+        logger.debug(
+            "PUT request to %s with data: %s and parameters: %s",
+            url,
+            data,
+            self._params,
+        )
+        kwargs.setdefault("timeout", 30)
+        params = kwargs.pop("params", {})
+        try:
+            res = requests.put(
```
antarcticrainforest, line 847: Nitpick: that should be post.

freva-client/src/freva_client/query.py
```diff
@@ -757,3 +830,51 @@ def _get(
...
+            res.raise_for_status()
+            return res
+        except KeyboardInterrupt:
+            pprint("[red][b]User interrupt: Exit[/red][/b]", file=sys.stderr)
+        except (
+            requests.exceptions.ConnectionError,
+            requests.exceptions.HTTPError,
+        ) as error:
+            msg = f"DELETE request failed with {error}"
+            logger.warning(msg)
```
antarcticrainforest, line 879: Same for the error here. The get method is in my opinion a little different cause a user can decide on what happens if a search has no results. But things might be a little different in delete or post scenarios. Maybe you could return the number of deleted/added objects. The users can decide on what to do if it's 0 but should be more than 0.

freva-rest/src/databrowser_api/endpoints.py
```diff
@@ -235,3 +243,78 @@ async def load_data(
         status_code=status_code,
         media_type="text/plain",
     )
+
+
+@app.put(
```
antarcticrainforest, line 248: Nitpck: this should be post.

freva-rest/src/databrowser_api/endpoints.py
```diff
@@ -235,3 +243,78 @@ async def load_data(
...
+    status_code=status.HTTP_202_ACCEPTED,
+    tags=["User Data"],
+)
+async def put_user_data(
+    username: str,
+    paths: Annotated[
+        Sequence[str], Query(description="Paths to the user's data to be added")
+    ],
+    facets: Annotated[
+        Mapping[str, str], Body(..., description="Facet key-value pairs for metadata")
```
antarcticrainforest, line 259: IMOH:  delte Facet

freva-rest/src/databrowser_api/endpoints.py
```diff
@@ -235,3 +243,78 @@ async def load_data(
...
     )
+
+
+@app.put(
+    "/api/databrowser/userdata/{username}",
+    status_code=status.HTTP_202_ACCEPTED,
+    tags=["User Data"],
+)
+async def put_user_data(
+    username: str,
```
antarcticrainforest, line 254: Do you need a username if you can take it from the token?

freva-rest/src/databrowser_api/endpoints.py
```diff
@@ -235,3 +243,78 @@ async def load_data(
...
+    }
+
+
+@app.delete(
+    "/api/databrowser/userdata/{username}",
+    status_code=status.HTTP_202_ACCEPTED,
+    tags=["User Data"],
+)
+async def delete_user_data(
+    username: str,
```
antarcticrainforest, line 298: Same as above about the token, do you need a username if you can take it from the token?

freva-rest/src/freva_rest/__init__.py
```diff
@@ -1,7 +1,7 @@
 import os
 from pathlib import Path
 
-__version__ = "2408.0.0"
+__version__ = "2410.0.0-beta2"
```
antarcticrainforest, line 4: You can bump the version to 2410.1.0

freva-rest/src/databrowser_api/core.py
```diff
@@ -887,3 +1071,426 @@ async def zarr_response(
...
+        Returns
+        ~~~~~~~
+        str
+            The CMOR-compliant time frequency.
+        """
+        if freq_attr in self.time_table.values():
+            return freq_attr
+        return self._timedelta_to_cmor_frequency(time_delta)
+
+    async def _get_metadata(self, file_name: os.PathLike[str]
```
antarcticrainforest, line 1360: This is not the right place for this method. I think we should add this to the client. I know it would be nice if this would work but you'd need to guarantee that that server has access to the data. In our case we would have to have the entire lustre mounted. Also the server would have to have read access to the data.

I think you want to place this method to the client.

antarcticrainforest (changes requested):
Ok first round done.

Looking, good so far. The only point that I have is the place where you collect the metadata. This needs to be done on the client side. We don't have to make sure that the server has read access to the data.

freva-rest/src/databrowser_api/endpoints.py
mo-dkrz: I'm good with changing this to `post` but can i ask what ur reason is? Since both methods are going to do a same job, in principal switching to `post` doesn't change anything in the functionality and final result. So no worries about technical part. But i read a bit about these CRUD methods and restful api principals, and it seems `put` is idempotent, whihc means, multiple same req has a same effect, or it might update if there is any changes. And in this case, add-user-data, user can add/update same data as many as wishes, without any effect. But from another point of view, i think u r right! because we call it specifically `add`-user-data, so it's makes more sense to call it `post`. i've already changed to `post`, but plz let me know what ur concern is regarding changing the method ....

antarcticrainforest: POST is used when the operation involves submitting data to be processed, which is the case when users are adding or indexing new data. This means the server is doing some processing (like adding the data to the index), and the data being sent may result in a new resource or trigger a specific action. 

If, on the other hand, the index location would be known then PUT would be appropriate. But since we cannot tell this I'd go for POST. Also, apache solr always uses POST, even for delete actions. 



If I think about it, you probably can make the delete methods also POST.

mo-dkrz:
Thanks @antarcticrainforest for the constructive reviews,
all request changes have been reflected and the pre-released version has been already operational on freva dev system.
if it's possible plz take a test on ur own machine or ur local env on levante for a random dataset to see the logs. cause I'm not totally happy about the logs so far.

how the cli and python syntax look like:

```bash
pip install git+https://github.com/FREVA-CLINT/freva-nextgen.git@add-user-data#subdirectory=freva-client
token=$(freva-client auth -u <k-number> --host freva.dkrz.de | jq -r .access_token)
freva-client databrowser user-data add \
    --path /a/dataset/path/on/ur/machine/or/levante/ \
    --facet project=cordex \
    --facet experiment=rcp85 \
    --facet model=mpi-m-mpi-esm-lr-clmcom-cclm4-8-17-v1 \
    --facet variable=tas \
    --host freva.dkrz.de \
    --access-token $token
```
and then delete:
```bash
freva-client databrowser user-data delete \
    --search-key project=cordex \
    --search-key experiment=rcp85 \
    --search-key model=mpi-m-mpi-esm-lr-clmcom-cclm4-8-17-v1 \
    --search-key variable=tas \
    --host freva.dkrz.de \
    --access-token $token
```

or alternatively on pytohn, syntax looks like:
```python
from freva-client import databrowser, authentication
token = authentication(k-number)
db = databrowser(host="freva.dkrz.de")
xarray_dataset = "/path/to/file.nc"
db.userdata("add", userdata_items=[xarray_dataset, "/path/tp/file2.nc"],metadat={})
db.userdata("delete")
```
how to find it on freva-dev instance:
https://www.freva.dkrz.de/databrowser/?flavour=user&user=k-number

but restful api endpoints now doesn't look pretty much applicable as initial point of user-data crawling! cuase it works as a bridge between client and server to transfer metadata from client to server

freva-rest/src/databrowser_api/core.py
```diff
@@ -1127,356 +1122,175 @@ def escape_special_chars(value: str) -> str:
...
+            "fs_type": self.fs_type
+        } | fwrites.get("facets", {})
         try:
-            user, paths, fwrites = args
-            self.validated_userdata = paths
-            self.fwrites |= {"user": user,
-                             "fs_type": self.fs_type} | cast(Dict[str, str], fwrites)
-            await self._ingest_user_data()
-
+            self.validated_user_metadata = user_metadata
```
antarcticrainforest, line 1281: I think you want to pass user_metdata as an argument rather than assigning it to a class attribute. The submit function takes the metadata list as an argument anyway. So you might just pass it through.

freva-rest/src/databrowser_api/core.py
```diff
@@ -1127,356 +1122,175 @@ def escape_special_chars(value: str) -> str:
...
-        """
-        for item in Path(path).rglob("*.*"):
-            if item.is_file() and any(
-                item.suffix == suffix for suffix in self.suffixes
-            ):
-                yield item
 
-    async def _process_paths_in_executor(
-            self, paths: List[Path], path_to_mongo: Path) -> None:
+    async def _process_user_metadata_in_executor(self,
```
antarcticrainforest, line 1174: Do we really need that function? I understand that it was there when you actually opened files to get the metadata. This could have been a potentially long process. But since we have all metadata aboard I think we don't need to run this in a different thread at all. Adding the data to solr and mongo should be fast.

freva-rest/src/databrowser_api/endpoints.py
```diff
@@ -245,72 +262,57 @@ async def load_data(
...
-        raise HTTPException(
-            status_code=status.HTTP_400_BAD_REQUEST,
-            detail=f"Invalid paths provided: {str(e)}",
+        await solr_instance.add_user_metadata(
+            current_user.preferred_username,  # type: ignore
+            validated_user_metadata,
+            facets=request.facets
         )
-    except Exception as e:  # pragma: no cover
+    except Exception as e:
```
antarcticrainforest, line 286: I think you want to log an exception here - in case someone comes and complains that the user data won't ingest so an admin could have a look

freva-client/src/freva_client/query.py
```diff
@@ -712,103 +745,132 @@ def _facet_search(
             constraints = data["primary_facets"]
         return {f: v for f, v in data["facets"].items() if f in constraints}
 
-    def add_user_data(
-        self, username: str, paths: List[str], facets: Dict[str, str]
+    def userdata(
+        self,
+        action: Literal["add", "delete"],
+        userdata_items: List[Union[str, xr.Dataset]] = [],
+        metadata: Dict[str, str] = {},
```
antarcticrainforest, line 752: This can have very confusing consequences for the users. Do you assign these mutable objects in the function signature on purpose? If so rename them to _userdata_items, _metadata to show that this is internal. Otherwise, assign them to None and do a `userdata_times = userdata_items or []` etc. later.

freva-client/src/freva_client/query.py
```diff
@@ -712,103 +745,132 @@ def _facet_search(
...
+                    "user_metadata": self.user_metadata,
+                    "facets": metadata,
+                }
+                result = self._post(url, data=self.payload_metadata, headers=headers)
+                pprint("[b][green]User data added successfully[green][b]")
+                if result is None:
+                    raise ValueError("Failed to add user data")
+            else:
+                raise ValueError("No metadata generated from the input data.")
+
```
antarcticrainforest, line 862: Just to let you know, users will create instances of this object. Because of that we should be careful with attributes like user_metadata, payload_metadata etc. They aren't for the users. So anything that is a utility in this class should start with a `_` to indicate to the users that it's not for direct access. In general, I think it might be wiser to outsource most of the functionality of used in this method (validation, processing, ...) to a different user data class that is instantiated and garbage collected in this method.

freva-client/src/freva_client/query.py
```diff
@@ -831,22 +893,19 @@ def _get(self, url: str, **kwargs: Any) -> Optional[requests.models.Response]:
             logger.warning(msg)
         return None
 
-    def _put(
+    def _post(
```
antarcticrainforest, line 896: I think, rather than adding extra methods that do kind of similar things, you could do one method that gets the request type to rule them all. So instead of `reuests.get, requests.put` etc you use `requests.Request("GET",...)` etc.

freva-client/src/freva_client/query.py
```diff
@@ -876,5 +937,205 @@ def _delete(self, url: str, **kwargs: Any) -> Optional[requests.models.Response]
...
+        user_data: Sequence[Union[str, xr.Dataset]],
+    ) -> Tuple[List[os.PathLike[str]], List[xr.Dataset]]:
+
+        validated_paths: List[os.PathLike[str]] = []
+        validated_xarray_datasets: List[xr.Dataset] = []
+
+        for data in user_data:
+            if isinstance(data, str):
+                path = Path(data)
+                if not path.exists():
```
antarcticrainforest, line 956: I think it would be kind of nice to have a functionality that can handle wild cards, so that something like this would work `~/data/*.nc`.

freva-client/src/freva_client/query.py
```diff
@@ -876,5 +937,205 @@ def _delete(self, url: str, **kwargs: Any) -> Optional[requests.models.Response]
...
+        if not validated_paths and not validated_xarray_datasets:
+            raise FileNotFoundError("No valid file paths or xarray datasets found.")
+
+        return validated_paths, validated_xarray_datasets
+
+    async def _process_user_data(self) -> None:
+        tasks: List[asyncio.Future[None]] = []
+        tasks.extend(
+            [
+                asyncio.ensure_future(
```
antarcticrainforest, line 972: Why did you decide on making this asyncio when the lib isn't async at all?

freva-client/src/freva_client/query.py
```diff
@@ -876,5 +937,205 @@ def _delete(self, url: str, **kwargs: Any) -> Optional[requests.models.Response]
...
+            for var in data_vars
+            if var not in coords
+            and not any(
+                term in var.lower() for term in ["lon", "lat", "bnds", "x", "y"]
+            )
+            and var.lower() not in ["rotated_pole", "rot_pole"]
+        ]
+        # TODO: we need to find out why we are not getting more than one variable!
+        if len(variables) != 1:
+            logger.warning("Only one data variable allowed, found: %s", variables)
```
antarcticrainforest, line 1129: I think, you can drop this constraint.

antarcticrainforest (changes requested):
Ok, seconde round done!

I've two concerns about the client:

- I think the gathering of metadata should happen in a helper class in directly in the databrowser class
- I am not sure whether we want to make the code async. At least not methods that are defined in the databrowser class. Remember, that users will use this class.

mo-dkrz:
@antarcticrainforest thanks for the second round. The update is now ready for review again.
The main change from the user’s perspective is the Python syntax update, enabling users to specify the `host` and `fail_on_error` parameters directly in the `userdata` method. On the other hand, wildcard paths are now supported for crawling.

```python
from freva-client import databrowser, authentication
import xarray as xr
token = authentication(k-number, host="freva.dkrz.de")
file_path = "/path/to/file.nc"
xarray_dataset = xarray.open_dataset("file_path")
databrowser.userdata("add", userdata_items=[xarray_dataset, "/path/tp/file2.nc"], metadat={}, host="freva.dkrz.de", fail_on_error=True)
databrowser.userdata("delete", host="freva.dkrz.de", fail_on_error=True)
```

freva-client/src/freva_client/query.py
```diff
@@ -733,26 +720,170 @@ def _facet_search(
...
+        userdata_items = userdata_items or []
+        metadata = metadata or {}
+        url = f"{this._cfg.userdata_url}"
+        token = this._auth.check_authentication(auth_url=this._cfg.auth_url)
+        headers = {"Authorization": f"Bearer {token['access_token']}"}
+        payload_metadata: dict[str, Collection[Collection[str]]] = {}
+
+        if action == "add":
+            user_data_handler = UserDataHandler(userdata_items)
+            if user_data_handler._user_metadata:
```
antarcticrainforest, line 824: Please don't access private attributes.

freva-client/src/freva_client/utils/databrowser_utils.py
```diff
@@ -177,3 +189,239 @@ def get_dirs(user: bool = True) -> Path:
...
+            time_freq = dset.attrs.get("frequency", "")
+            data_vars = list(map(str, dset.data_vars))
+            coords = list(map(str, dset.coords))
+            try:
+                times = dset["time"].values[:]
+            except (KeyError, IndexError, TypeError):
+                times = np.array([])
+
+        except Exception as error:
+            logger.warning("Failed to open data file %s: %s", str(path), error)
```
antarcticrainforest, line 386: That should be logging an error

freva-client/src/freva_client/utils/databrowser_utils.py
```diff
@@ -177,3 +189,239 @@ def get_dirs(user: bool = True) -> Path:
...
+        batch = []
+        for item in items:
+            if isinstance(item, (Path, xr.Dataset)):
+                batch.append(item)
+                if len(batch) >= self._batch_size:
+                    yield cast(Union[List[os.PathLike[str]], List[xr.Dataset]], batch)
+                    batch = []
+
+        if batch:
+            yield cast(Union[List[os.PathLike[str]], List[xr.Dataset]], batch)
```
antarcticrainforest, line 321: Do you really need this _batch_data metho? Isn't the thread pool taking care of it already?

freva-client/src/freva_client/utils/databrowser_utils.py
```diff
@@ -177,3 +189,239 @@ def get_dirs(user: bool = True) -> Path:
...
+
+    This class is used for processing user data and extracting metadata
+    from the data files.
+    """
+    def __init__(self, userdata_items: List[Union[str, xr.Dataset]]) -> None:
+
+        self._suffixes = [".nc", ".nc4", ".grb", ".grib", ".zarr", "zar"]
+        self._batch_size = 150
+        self._lock = Lock()
+        self._user_metadata: List[Dict[str, Union[str, List[str], Dict[str, str]]]] = []
```
antarcticrainforest, line 210: self.user_metadata

antarcticrainforest (changes requested):

mo-dkrz:
@antarcticrainforest i resolved all change reqs. Logs are still don't look satisfying to me ...

mo-dkrz:
@antarcticrainforest now logs look better and we have a stage to validate Solr before post ...

freva-client/src/freva_client/utils/databrowser_utils.py
```diff
@@ -203,138 +200,82 @@ class UserDataHandler:
...
-                        logger.warning(f"File path does not exist: {data}")
-                        continue
+            if isinstance(data, (str, Path)):
+                path = Path(data).expanduser().absolute()
+                if path.is_dir():
+                    validated_paths.extend(self._gather_files(path))
+                elif path.is_file() and path.suffix in self._suffixes:
                     validated_paths.append(path)
+                else:
+                    validated_paths.extend(
```
antarcticrainforest, line 234: You can mix lists and interators? Interesting

freva-rest/src/databrowser_api/core.py
```diff
@@ -1208,9 +1277,22 @@ async def add_user_metadata(
...
         )
+        if self.total_ingested_files == 0:
+            status_msg = (
+                "No data was added to the databrowser. "
+                "(No files ingested into Solr and MongoDB)"
+            )
+        else:
+            status_msg = (
+                f"Your data has been successfully added to the databrowser. "
+                f"(Ingested {self.total_ingested_files} files into Solr and MongoDB)"
```
antarcticrainforest, line 1292: You can just have it saying: f"{self.total_ingested_files} have been successfully added to the databrowser."

freva-rest/src/databrowser_api/core.py
```diff
@@ -1115,19 +1170,33 @@ async def _ingest_user_metadata(self,
...
-                await self._add_to_solr(user_metadata_batch)
-                await self._insert_to_mongo(user_metadata_batch)
-                user_metadata_batch = []
-
-        if user_metadata_batch:
-            await self._add_to_solr(user_metadata_batch)
-            await self._insert_to_mongo(user_metadata_batch)
+            if not uri and not file_path:
+                continue
+            is_duplicate = await self._is_query_duplicate(str(uri), str(file_path))
```
antarcticrainforest, line 1196: You could also collect the number of duplicates and then add this to the message that n entries were not added because they were duplicates.

freva-rest/src/databrowser_api/endpoints.py
```diff
@@ -235,3 +261,65 @@ async def load_data(
...
+    try:
+        validated_user_metadata = await solr_instance._validate_user_metadata(
+            request.user_metadata
+        )
+        status_msg = await solr_instance.add_user_metadata(
+            current_user.preferred_username,  # type: ignore
+            validated_user_metadata,
+            facets=request.facets
+        )
+    except Exception as e:
```
antarcticrainforest, line 287: Could you distinguish at least for validation errors? That way we get a 422 on failed validations. The rest can be 500

antarcticrainforest (changes requested):
Tow minor comments.

antarcticrainforest (approved):
LGTM

PR #92, opened 2024-10-24
qf of the issue of opening the downloaded intake catalog

mo-dkrz:
This pull request includes a quick updates to the `core.py` file to fix the issue of opening of the intake catalog that we download from `api/databrowser/intake_catalogue/{flavour}/file` endpoint

* Added the `format` field to the list of facets in the `facet_hierarchy` method
* Changed the `column_name` from `uri` to `file` in the `_create_intake_catalogue` method to correctly reference the asset column name

### DIFF
--- a/freva-rest/src/databrowser_api/core.py
+++ b/freva-rest/src/databrowser_api/core.py
@@ -559,7 +559,7 @@ class SolrSearch:
                 }
                 for v in facets
             ],
-            "assets": {"column_name": "file", "format_column_name": "format"},
+            "assets": {"column_name": self.uniq_key, "format_column_name": "format"},
             "id": "freva",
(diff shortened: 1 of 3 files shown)


freva-rest/src/databrowser_api/core.py
```diff
@@ -558,7 +559,7 @@ async def _create_intake_catalogue(self, *facets: str) -> IntakeType:
                 }
                 for v in facets
             ],
-            "assets": {"column_name": "uri", "format_column_name": "format"},
+            "assets": {"column_name": "file", "format_column_name": "format"},
```
antarcticrainforest, line 562: Does this always have to be file or should this be `self.uniq_key`? Depending of what the user asked for?

antarcticrainforest (approved):
LGTM

PR #96, opened 2024-11-04
Refactor tests

antarcticrainforest:
This PR aims to make the tests a little easier to comprehend and to address some bugs that were arising due to opened and closed MongDB/Redis connections. @mo-dkrz


tests/test_databrowser_restapi.py
mo-dkrz, line 1: neat and organized, perfect!

mo-dkrz (commented):
Overall look good to me, but, as a minor comment what do you think to transfer the whole tests in `tests/test_userdata.py` into `tests/test_databrowser_restapi.py` since both are related to each other? Or we I can take care of this part after merging. 

Also one off-topic point which looked important to me was calling `self._get_solr_fields()` in config file of freva_rest. What do you think to call it directly in `core.py`

tests/test_databrowser_restapi.py
```diff
@@ -0,0 +1,364 @@
...
+        params={"product": "cmip", "translate": "true"},
+    ).status_code
+    res3 = requests.get(
+        f"{test_server}/api/databrowser/data_search/cmip6/uri",
+        params={"activity_": "cmip"},
+    ).status_code
+    assert res1 == res2 == res3 == 422
+
+
+def test_no_mongo_parameter_insert(test_server: str) -> None:
```
mo-dkrz, line 281: I think for making the coverage full, we need a test like this for `_delete_from_mongo`. i just copied your `test_no_mongo_parameter_insert` and made it based on that. 

```python
def test_no_mongo_parameter_delete(test_server: str) -> None:
    """Test the failed deletion of data into the mongodb."""
    from freva_rest.rest import server_config
    token = auth["access_token"]
    with MongoClient(server_config.mongo_url) as mongo_client:
        collection = mongo_client[server_config.mongo_db]["user_data"]
        collection.drop()
        assert collection.find_one() is None
    env = os.environ.copy()
    env["MONGO_HOST"] = "foo.bar.de"
    with mock.patch.dict(os.environ, env, clear=True):
        server_config.power_cycle_mongodb()
        res1 = requests.delete(
            f"{test_server}/api/databrowser/userdata",
            json={},
            headers={"Authorization": f"Bearer {auth['access_token']}"},
        ).status_code
        assert response.status_code == 202
    server_config.power_cycle_mongodb()
    with MongoClient(server_config.mongo_url) as mongo_client:
        collection = mongo_client[server_config.mongo_db]["user_data"]
        assert collection.find_one() is None
```

mo-dkrz (changes requested):
could you try this test, cause it might cover the exception of the delete monog

tests/test_userdata.py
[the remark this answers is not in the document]
antarcticrainforest: Those tests were actually only meant for testing data queries. I've updated the name to make it a little more clear.

tests/test_databrowser_restapi.py
antarcticrainforest: I've taken the liberty to make converagerc not to consider those lines because we don't do anything special other than logging an error if an issue occurs.

mo-dkrz (approved):

PR #104, opened 2024-11-15
Endpoint boilerplate

antarcticrainforest:
I careated a new branch that (new-tool-api) that we can use as dev branch to merge into.

I think we should merge more often to keep a better overview.

This PR aims at getting skeloton for the restAPI. We can discuss this in the meeting.


freva-rest/src/tool_api/endpoints.py
```diff
@@ -0,0 +1,45 @@
+from typing import Dict
```
mo-dkrz, line 1: what is this endpoints and the other in `freva-rest/src/freva_rest/tool_api/endpoints.py`? could u plz concatenate them in one?

freva-rest/src/tool_api/endpoints.py
```diff
@@ -0,0 +1,45 @@
...
+                        summary="",
+                        added="2024-11-13",
+                        command="fakecommand",
+                        binary="fakebinary",
+                        output_type="data"
+    )
+    output = tool.model_dump()
+    return JSONResponse(content=output)
+
+@app.post("/api/tool/add",
```
mo-dkrz, line 29: what do u mean  by `add`? submit the job you meant? `Add a tool to the MongoDB` for what? did u considered auth?

freva-rest/src/tool_api/endpoints.py
```diff
@@ -0,0 +1,45 @@
...
+          tags=["Tool Addition"])
+async def add_tool(tool: ToolAbstract,
+                   current_user: TokenPayload = Depends(auth.required)
+) -> Dict[str, str]:
+    """Add a tool to the MongoDB"""
+    # TODO: Implement method
+    return {"status": "This method is not implemented yet!"}
+
+@app.delete("/api/tool/del")
+async def delete_tool(tool_id:int,
```
mo-dkrz, line 40: the same as above

freva-rest/src/freva_rest/tool_api/tool_abstract.py
```diff
@@ -0,0 +1,302 @@
...
+    type: Literal["File"] = "File"
+
+    
+    @field_validator("default")
+    @classmethod
+    def assert_is_file(cls, path: Path):
+        if path and not path.is_file():
+            raise ValueError("Default value must be a file!")
+        return path
+
```
mo-dkrz, line 215: how about Solrfield? it;s the most important one i think

freva-rest/src/freva_rest/tool_api/parse_config.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Raises:
+        AssertionError: Raised when input_file does not exist or does not have a valid TOML file extension.
+        IOError: Raised when file cannot be opened in read-only mode. 
+        tomllib.TOMLDecodError: Raised when TOML file cannot be parsed by tomllib.
+    """
+    if isinstance(input_file, str): input_file = Path(input_file)
+    valid_toml_file_extensions = [".toml"]
+    assert input_file.exists(), f"Input file {input_file} cannot be found. Please make sure it exists and that you have read permissions for it." 
+    assert input_file.is_file() and input_file.suffix.lower() in valid_toml_file_extensions, f"Path {input_file} does not point to a recognised TOML file."
```
mo-dkrz, line 29: ahhh, since these checks are part of the function's core validation logic, not debugging, i think you want to get them through this way:

```python
if not input_file.exists():
    raise FileNotFoundError(f"Input file {input_file} cannot be found. Please make sure it exists and that you have read permissions for it.")
    
if not input_file.is_file() or input_file.suffix.lower() not in valid_toml_file_extensions:
    raise ValueError(f"Path {input_file} does not point to a recognised TOML file.")
```

freva-rest/src/tool_api/endpoints.py
felio92: You are right, I removed the superfluous folder now

freva-rest/src/freva_rest/tool_api/tool_abstract.py
felio92: I defined a field called `DataField` that takes over this role. I wanted to make it more generic, as the backend for the databrowser might not be Solr.

freva-rest/src/tool_api/endpoints.py
felio92: Add the tool to the MongoDB is the idea here. See `freva-rest/src/freva_rest/tool_api/endpoints.py` for an updated version of this. The idea is that this is the endpoint to post the TOML config of the tool to. Validation of the TOML file would be done server-side and depending if the user is authenticated and the TOML config is valid, endpoint would return a 202 accepted message or one of a list of errors.

freva-rest/src/tool_api/endpoints.py
felio92: This is the deletion analog of the `add` endpoint, allowing existing tools to be deleted from the MongoDB.

freva-rest/src/freva_rest/tool_api/parse_config.py
felio92: Thanks, this file definitely needs more work. I will take your suggestions on board!

PR #111, opened 2024-11-22
Add STAC STATIC and BBOX

mo-dkrz:
Hi @antarcticrainforest 
This PR implements initial STAC API integration in the Databrowser and replaces aiohttp with httpx across all avaiable CRUDs in core of databrowser
But this is only producing the STAC API from user searches in databroswer, not considering STAC-API service as the search system of databrowser.
I liked to demonstrate the prototype of the STAC integration on the dev machine, but since we needed the `bbox` field on Solr and on the Solr instance of dev machine, currently we don't have it, so we need to first get the crawler MR done, crawl it, and then we will be able to play around with the pre-release version.

But first and foremost, we need to get [the freva-service PR](https://github.com/FREVA-CLINT/freva-service-config/pull/27) done to be able to keep up with data crawler. Could you please first review that? As soon as you finished the review of freva-service PR, I submit another PR in Freva to have the `freva-dev` docker image which I need for the data Crawler pipeline.


antarcticrainforest:
Can I already look at this?

antarcticrainforest (changes requested):
I've had a first look at this. But I still need check and try out how this works in practice. This is only going to happen in February though.

I have a few general comments that I just post here:

1. container: Instead of using a private container I think stac server service can be integrated into the existing freva-rest server container. Since we already start various services in the container, I think it doesn't hurt to start another one.

2. Please make sure that any of the infrastructure development/deployment is in line with being able to setup the whole server with conda-forge.

3. I think the stac catalogue creation should happen in a class on it's own. That way the Solr class becomes less busy and if we add  at some point in the future the functionality of adding various backends it'll be easier to maintain and understand.

4. Also think that other institutions might be potentially using this. You can either use generic sources like freva version bla bla as we do in the intake catalogue creation or use more specific metadata which has to be passed via config (I am not in favour for that option as it makes the already existing configuration more complex).

mo-dkrz:
Dear @antarcticrainforest static catalog also has been added to this PR. It means we can get the STAC catalog as a `tar.gz` file from Freva along with a STAC collection service. Now the the prototype on dev machine is operational. You can experience both bbox and STAC catalogs there. Also front-end is somehow done. But for redirecting to dynamic STAC API to STAC browser we still have some minor issues which I'm trying to find a solution for it. So it means it redirects to `localhost:8085` which `localhost` has to be replaced with dev machine host address to see the result. I might have one or two more commit here left to fix this.

freva-client/src/freva_client/query.py
```diff
@@ -75,6 +77,23 @@ class databrowser:
...
         covered. ``file`` will only return files where the entire time
         period is contained within `one single` file.
+    bbox: str, default: ""
+        Special search facet to refine/subset search results by spatial extent.
+        This can be a string representation of a bounding box or a WKT polygon.
+        Valid strings are ``min_lon,max_lon by min_lat,max_lat`` for bounding
+        boxes and Well-Known Text (WKT) format for polygons.
+
+        .. note:: Longitude values must be between -180 and 180, latitude values
+                    between -90 and 90.
```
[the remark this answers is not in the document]
antarcticrainforest:
```
  min_lon, max_lon, min_lat, max_lat = bounds

    def normalize_lon(lon):
        # Convert a longitude from 0-360 to -180 to 180 if necessary.
        if lon > 180:
            return lon - 360
        if lon < -180:
            return lon + 360
        return lon

    # Normalize longitudes
    new_min_lon = normalize_lon(min_lon)
    new_max_lon = normalize_lon(max_lon)
```

freva-data-portal-worker/pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
 "watchfiles",
 "xarray",
 "xpublish",
-"zarr",
+"zarr==2.18.2",
```
antarcticrainforest, line 40: zarr<3?

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -261,21 +266,55 @@ async def stac_collection(
...
+        background_tasks.add_task(run_stac_creation)
+
+        redirect_url = (
+            f"{server_config.stacbrowser_host}"
+            f"/collections/{collection_id}"
+        )
+        # IMPORTANT: wait for the background task to start.
+        # Otherwise the client will get a 404 error and
+        # has to reload the page to see the STAC collection.
+        await asyncio.sleep(1)
```
antarcticrainforest: Is there a better way to determine if the background task has started?

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -261,21 +266,55 @@ async def stac_collection(
...
-    collection_id = f"freva-{str(uuid.uuid4())}"
-    background_tasks.add_task(solr_search.init_stac_collection, request, collection_id)
-    return {
-        "status": (
-            f"STAC catalog creation initiated successfully. "
-            f"To see the collection, use this link: "
-            f"{request.base_url}/stac/{collection_id}"
+
+    collection_id = f"Dataset-{(f'{flavour}-{str(uuid.uuid4())}')[:18]}"
+    if stac_dynamic:
```
antarcticrainforest, line 277: For my taste this if block is too big. Could you add the code that is withing this if block into a function? Or at least the `run_stac_creation` function.

antarcticrainforest (changes requested):

mo-dkrz:
The only part that is left here is, adding the opensearch or elasticsearch configuration in conda recipe for both container and non container deployment purposes

freva-rest/src/freva_rest/databrowser_api/core.py
```diff
@@ -658,16 +703,73 @@ def adjust_time_string(
...
+            have any overlap with the query extent. ``file`` will only return files
+            where the entire spatial extent is contained by the query geometry.
+
+        Raises
+        ------
+        ValueError: If parsing the coordinates failed or if bbox_select is invalid.
+        """
+        if not bbox:
+            return []
+        bbox = "".join(part.strip() for part in bbox.split())
```
antarcticrainforest, line 746: It would be better if the code would expect bbox to be a list or tuple for that matter of four entries e.g [0,50, 100, 120]

antarcticrainforest (changes requested):
I am not a fan of the bbox definition 0, 20 by 100, 120. To me it would be better if this would be a list of four numerals [0, 20, 100, 120]. 

Could you change the code accordingly. Could you also add an example to the restAPI docs that uses the bbox?

antarcticrainforest (approved):
LGTM!

mo-dkrz:
@antarcticrainforest I think it's ready to go, but after merging freva-service [PR](https://github.com/FREVA-CLINT/freva-service-config/pull/29) remove redundant configuration and  #143 to get green CI

antarcticrainforest:
I quickly tested the cli:

```
freva-client databrowser stac-catalogue dataset=cmip6-fs --host localhost:7777                                                                                                                       (freva-dev) 
[2025-04-01T10:57:47] ERROR    freva-client: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'NoneType' 
```

The rest seems to be fine

antarcticrainforest (changes requested):

antarcticrainforest (approved):
I haven't checked why the tests are failing. But once this is solved I think you can merge it.

After you've done so, could you already create a new release. I am trying to do my best to get the deployment ready.

mo-dkrz:
> I haven't checked why the tests are failing. But once this is solved I think you can merge it.
> 
> After you've done so, could you already create a new release. I am trying to do my best to get the deployment ready.

In the latest changes, two main changes happened.

1. We've completely removed the `pystac` library from our dependencies. Since the library was primarily used to generate a few JSON files conforming to the STAC spec, I decided to replace it with four lightweight custom classes that handle this functionality internally. Maintaining external dependencies like `pystac` can become so expensive for us, and we anticipate needing to make infra-specific adjustments to this feature in the future. Rewriting these components ourselves allows for better control, easier maintenance, and remvoes the dependency overhead. But in the meanwhile still the generated STAC catalogs pass validation via `stac-check` which I designed one test for it.

2. Given the STAC spec's modular structure, we need a way to package and deliver all components of a STAC catalog together to user via downlaoding. For this, we've moved from using the `tar.gz` format to `zip`. There are two main reasons behind this change. First, zip is more accessible for Windows users. Second adn the main reason, zip allows us to generate and stream the JSON files on the fly without needing to know their final size ahead of time. From other side, tar archives require the file size up front, which makes streaming mem inefficient.

@antarcticrainforest could you please have a look again.

mo-dkrz:
by the way the problem of freva-client has been fixed as well:
```python
>>> from tempfile import mktemp
... temp_path = mktemp(suffix=".zip")
... 
... from freva_client import databrowser
... db = databrowser(dataset="cmip6-hsm", host="localhost:7777")
... db.stac_catalogue(filename=temp_path)
... print(f"STAC catalog saved to: {temp_path}")
... 
Downloading the STAC catalog started ...
STAC catalog saved to: /var/folders/99/_rg5v0pn0sq81l0rr5g0z20h0000gn/T/tmpecf3t_fz.zip
```

cli:

```bash
❯ freva-client databrowser stac-catalogue --host localhost:7777
Downloading the STAC catalog started ...
STAC catalog saved to: /Users/mo/dev/20250331/freva-nextgen/stac-catalog-Dataset-freva-61c5fb3b-c8d-file.zip (size: 0.24 MB). Or simply download from: http://localhost:7777/api/freva-nextgen/databrowser/stac-catalogue/freva/file?multi-version=False
```

mo-dkrz:
@antarcticrainforest To play around with the downloaded STAC static catalogs, we should use the `pystac`. it seems `pystac-client` is designed more for the STAC-API endpoints.

https://stacspec.org/en/tutorials/1-read-stac-python/

freva-client/src/freva_client/query.py
```diff
@@ -423,25 +423,25 @@ def intake_catalogue(self) -> intake_esm.core.esm_datastore:
 
     def stac_catalogue(
         self,
-        filename: Optional[Union[str, Path]] = None,
+        filename: Optional[Union[str, Path, None]] = None,
```
antarcticrainforest, line 426: That should be the same as `Optional[Union[str, Path]]`

antarcticrainforest (approved):
Just a small type annotation thing! Once fixed LGTM!

PR #118, opened 2024-12-05
Introduce an add tool logic

antarcticrainforest:
This PR aims at creating endpoints where users can add their tools to the central system. Users should also be able to change the visibility of their tools if they wouldn't like to clutter the available tools for them. I also figured that it would be good to have an option where users can make a tool public, that is available to other users. In this case the tool is given to to the admin user afterwards the user can't change the tool anymore (visibility etc0.

I've also already added configuration items for running the tool. So ideally when you start implementing the execute logic you can assume that the `Config` class has all information.

I hope this PR makes sense, let me know.


freva-rest/src/freva_rest/tool_api/db.py
```diff
@@ -324,12 +358,7 @@ class File(BaseParam):
     """
 
     type: Literal["path"]
```
felio92, line 360: Probably should be `Literal["range"]`

felio92:
Hi @antarcticrainforest ,
sorry, I just managed to go through the changes today. It looks good to me, and I think I understand the thought behind the changes.

PR #120, opened 2024-12-05
correct the freva-rest docker env variables

mo-dkrz:
In this PR we want to correct the most of environment variables of freva-rest Dockerfile. Also adding there were some small issues related to the Solr and Mongo that via this PR they all have been fixed.


mo-dkrz:
Since there is no [`numcodecs`](https://pypi.org/project/numcodecs/#files) pre-built wheel for linux/arm64, it's building the wheel on the CI and it's not time-efficient at all. So since adding ARM platform makes our ci 10 times slower for only producing image for arm64 arch, the solution would be caching the pre-built numcodecs  on our CI and re-use it again for other build image runs ahead. Or we can wait this [PR](https://github.com/zarr-developers/numcodecs/pull/563) be merged.

mo-dkrz:
@antarcticrainforest now it's ready for review. The issue of long installation of `numcodecs` on ARM64 has been fixed as well.

antarcticrainforest (approved):
LGTM!

PR #133, opened 2025-02-27
Introduce Freva backend agnosticism approach 

mo-dkrz:
This PR introduces a way to switch Freva’s default search backend from Solr to any available RDBMS or search engine backend using a simple configuration. If no secondary backend is set in the environment or `api_config.toml`, Solr will be used as a fallback. This makes Freva backend-agnostic, allowing it to work with different search backends.

Performance tests on STAC's native backends, pgSTAC and OpenSearch/ElasticSearch, show that it performs as well as Solr, with pgSTAC even performing better.

left:
Since this feature is needed for next week, Production tests on live STAC instances are still left but will be completed in the coming days.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "dev-env/config"]
 	path = dev-env/config
 	url = https://github.com/FREVA-CLINT/freva-service-config.git
+	branch = opensearch-stac
--- a/freva-rest/src/freva_rest/api_config.toml
+++ b/freva-rest/src/freva_rest/api_config.toml
@@ -112,3 +112,102 @@ client_id = "freva"
 # Secret key associated with the client ID. Keep this secure as it is used to
 # authenticate the API to the OIDC provider.
 client_secret = ""
+
+
+### IMPORTANT: If you uncomment the whole configuration below, you can use 
+## the secondary-backend for connecting to all kind of Rational Database
+## Management System (RDBMS) backend and search engine (SE) backend like
+## Elasticsearch or OpenSearch. You can uncomment one of the following
+## sections based on the backend you want to connect to.
+
+# # PostgreSQL Configuration
+
+# [secondary-backend]
+# # (OPTIONAL)Secondary Backend configuration for connecting to
+# # all kind of Rational Database Management System (RDBMS) backend and 
+# # search engine (SE) backend like Elasticsearch or OpenSearch.
+# # You can uncomment one of the following sections based on the 
+# # backend you want to connect to.
+
+# # In the following examples we generally consider the native 
+# # backends of STAC, pgSTAC and ES and OS as use cases.
+
+# type = "RDBMS"
+# dns = "postgresql+asyncpg://pgstac:secret@localhost:5432/postgis"
+# table = "pgstac.items"
+# pagination_column = "id" # IMPORTANT: specify the column name for pagination
+# limit_offset = "LIMIT :limit OFFSET :offset" # IMPORTANT: One can specify the limit and offset variables based on the database-specific syntax
+
+# [secondary-backend.lookuptable]
+# # secondary-backend.lookuptable section defines the lookup table 
+# # mapping between secondary backend fields and the data search params
+# # used in the Freva-API. 
+# # This allows the API to query secondary backend as backend of the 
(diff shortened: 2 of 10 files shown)


mo-dkrz:
@antarcticrainforest 

to test it, please:

```bash
docker compose -f dev-env/docker-compose.yml up -d --remove-orphans
conda create -n freva-nextgen-swap-back cfgrib
conda activate freva-nextgen-swap-back
make install
python dev-env/ingest_stac_metadata.py
```
and then uncomment [theses lines](https://github.com/FREVA-CLINT/freva-nextgen/blob/swap-freva-back/freva-rest/src/freva_rest/api_config.toml#L125-L181) for postgress or [these lines](https://github.com/FREVA-CLINT/freva-nextgen/blob/swap-freva-back/freva-rest/src/freva_rest/api_config.toml#L186-L213) for opensearch and also add the same configuration into the [dev-env/api_config.toml](https://github.com/FREVA-CLINT/freva-nextgen/blob/swap-freva-back/dev-env/api_config.toml); then

```bash
python run_server.py --config dev-env/api_config.toml --debug --dev -p 7777 -f
```
and then you can query from the backend you did the setup for!

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "dev-env/config"]
 	path = dev-env/config
 	url = https://github.com/FREVA-CLINT/freva-service-config.git
+	branch = opensearch-stac
```
antarcticrainforest, line 4: Shall we merge this upstream branch first?

freva-rest/src/freva_rest/api_config.toml
```diff
@@ -112,3 +112,102 @@ client_id = "freva"
...
+
+### IMPORTANT: If you uncomment the whole configuration below, you can use 
+## the secondary-backend for connecting to all kind of Rational Database
+## Management System (RDBMS) backend and search engine (SE) backend like
+## Elasticsearch or OpenSearch. You can uncomment one of the following
+## sections based on the backend you want to connect to.
+
+# # PostgreSQL Configuration
+
+# [secondary-backend]
```
antarcticrainforest, line 125: Why did you call this secondary backend? What's the primary. Also, why is this commented? 

Why you don't just call it `search-backend` or something else and merge this section with the solr section?

freva-rest/src/freva_rest/databrowser_api/core.py
```diff
@@ -75,6 +78,72 @@ async def wrapper(*args: Any, **kwargs: Any) -> asyncio.Task[Any]:
...
+    Returns
+    -------
+    Callable
+        Decorated function that handles secondary backend operations
+    """
+    def decorator(func: Callable) -> Callable:
+        @wraps(func)
+        async def wrapper(self, *args, **kwargs):
+            if self.backend_type in ("RDBMS", "SE"):
+                start_method_key = 'offset' if self.backend_type == 'RDBMS' else 'from'
```
antarcticrainforest, line 104: I really don't like this sort of comprehensions. Please write them in multiple lines.

freva-rest/src/freva_rest/databrowser_api/core.py
```diff
@@ -445,6 +549,240 @@ def __init__(
...
+                            if f"facet_{field}" in response["aggregations"]:
+                                buckets = response["aggregations"][f"facet_{field}"]["buckets"]
+                                results["facets"][translated_field] = [
+                                    item for bucket in buckets if bucket["key"]
+                                    for item in [str(bucket["key"]), int(bucket["doc_count"])]
+                                ]
+            return results
+
+        except Exception as e:
+            raise HTTPException(status_code=500, detail=f"[SE]: Search failed: {str(e)}")
```
antarcticrainforest, line 783: That's a bad idea. You want to log error messages and return a generic message.

antarcticrainforest (changes requested):
I don 't really like this current design pattern.

In order to implement this you want to refactor the current code. Create a base class for the search back ends that defines a get/query and a push/post/put method. Those methods then should interact with the search backend.

The translations and other things should be defined in this class. You can also define this in a new file.

Generally speaking if you find yourself writing complex code I think it's time to refactor the whole code.

PR #143, opened 2025-03-28
Improve logging and data-dirs.

antarcticrainforest:
This is a small PR that aims at unifying log and data output. The idea is that we can only use `-v data:/var/data` `-v logs:/var/log` to have the data and the logs of all services persist.

Note that I didn't for a unified folder such as `-v my-container-data:/var/freva-rest-server` and under `/var/freva-rest-server/` we have a `data` and `log` dir because it would make things even more easy but it would also break the volume structure of the current freva deployments, and backwards portability  is key.

### DIFF
--- a/freva-rest/docker-scripts/entrypoint.sh
+++ b/freva-rest/docker-scripts/entrypoint.sh
@@ -19,20 +19,23 @@ init_mongodb() {
     for env in API_MONGO_USER API_MONGO_PASSWORD; do
         check_env $env mongoDB
     done
-    /bin/bash /opt/conda/libexec/freva-rest-server/scripts/init-mongo
+    mkdir -p /var/data/mongodb /var/log/mongodb
+    API_DATA_DIR=/var/data/mongodb\
+        API_LOG_DIR=/var/log/mongodb /bin/bash /opt/conda/libexec/freva-rest-server/scripts/init-mongo
     log_info "Starting MongoDB with authentication..."
     /opt/conda/bin/mongod \
         -f /opt/conda/share/freva-rest-server/mongodb/mongod.yaml \
         --auth \
-        --cpu > /logs/mongodb.log 2>&1 &
+        --cpu 1> /dev/null 2> /var/log/mongodb/mongodb.err.log &
 }
 
 init_solr() {
     log_service "Initialising Solr"
     ulimit -n 65000 || echo "Warning: Unable to set ulimit -n 65000"
-    /bin/bash /opt/conda/libexec/freva-rest-server/scripts/init-solr
+    mkdir -p /var/data/solr /var/log/solr
+    API_DATA_DIR=/var/data/solr /opt/conda/libexec/freva-rest-server/scripts/init-solr
     log_info "Starting solr service"
-    nohup /opt/conda/bin/solr start -force > /logs/solr.log 2>&1 &
+    nohup /opt/conda/bin/solr start -force -Dsolr.log.dir=/var/log/solr 1> /dev/null 2> /var/log/solr/solr.err &
     SOLR_PORT=${API_SOLR_PORT:-8983}
     timeout 60 bash -c 'until curl -s http://localhost:'"$SOLR_PORT"'/solr/admin/ping;do sleep 2; done' || {
             echo "Error: Solr did not start within 60 seconds." >&2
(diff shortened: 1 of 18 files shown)


antarcticrainforest:
@mo-dkrz I've while I was add it I decided to build in a little feature.

When I played with the container I used the production keycloak instance and noticed that the server start got stuck because it could not connect to the keycloak. In case this ever happens I decided to override the auth method so that it checks whether or not keycloak is available and if not it returns a 503 for endpoints that need auth.

As a bonus: with this new function/class we would also be able to dynamically protect endpoints  by auth. Which might come in handy.

freva-rest/docker-scripts/entrypoint.sh
```diff
@@ -42,18 +66,22 @@ init_solr() {
...
     for env in MYSQL_USER MYSQL_PASSWORD MYSQL_ROOT_PASSWORD MYSQL_DATABASE; do
         check_env $env MySQL
     done
-    /bin/bash /opt/conda/libexec/freva-rest-server/scripts/init-mysql
+    mkdir -p /var/data/mysqldb /var/log/mysqldb
+    API_DATA_DIR=/var/data/mysqldb \
+        LOG_DIR=/var/log/mysqldb /opt/conda/libexec/freva-rest-server/scripts/init-mysql
     log_info "Starting mysql service"
-    nohup /opt/conda/bin/mysqld --user=$(whoami) > /logs/mysqld.log 2>&1 &
+    nohup /opt/conda/bin/mysqld --user=$(whoami) --no-defaults --datadir=/var/data/mysqldb > /var/log/myslqdb/mysqld.log 2>&1 &
```
mo-dkrz, line 84: typo which doesn't allow mysql starts: please change `/var/log/myslqdb/mysqld.log` to `/var/log/mysqldb/mysqld.log`

Also another comment here. MySQL is not accessible from outside of container on port `3306` even when it starts, which it seems it has to bind to all interface instead of it's default which is loop interface.

mo-dkrz (changes requested):
All looks good now!
1. only about REDIS, to give a nice experience to our future users, as a not very important comment, I think it would be better to define `check_env` in redis to inform necessary variables when it's enabled. For example based on one experiment, without defining the `API_REDIS_PASSWORD` I faced the following error on connecting to redis host from out of container:

```bash
$ docker exec -it <container-id> redis-cli -h localhost -p 6379 ping
PONG
$ redis-cli -h 0.0.0.0 -p 6379 ping
(error) DENIED Redis is running in protected mode because protected mode is enabled and no password is set for the default user. In this mode connections are only accepted from the loopback interface. If you want to connect from external computers to Redis you may adopt one of the following solutions: 1) Just disable protected mode sending the command 'CONFIG SET protected-mode no' from the loopback interface by connecting to Redis from the same host the server is running, however MAKE SURE Redis is not publicly accessible from internet if you do so. Use CONFIG REWRITE to make this change permanent. 2) Alternatively you can just disable the protected mode by editing the Redis configuration file, and setting the protected mode option to 'no', and then restarting the server. 3) If you started the server manually just for testing, restart it with the '--protected-mode no' option. 4) Set up an authentication password for the default user. NOTE: You only need to do one of the above things in order for the server to start accepting connections from the outside.
```
2. As a minor typo, `check_env` shows `>&2` in the message:
```bash
[2025-03-29 17:50:10] [ERROR] In order to set up mongoDB you must set the '$API_MONGO_USER' environment variable. >&2
```

mo-dkrz:
> @mo-dkrz I've while I was add it I decided to build in a little feature.
> 
> When I played with the container I used the production keycloak instance and noticed that the server start got stuck because it could not connect to the keycloak. In case this ever happens I decided to override the auth method so that it checks whether or not keycloak is available and if not it returns a 503 for endpoints that need auth.
> 
> As a bonus: with this new function/class we would also be able to dynamically protect endpoints by auth. Which might come in handy.

I think here, when the `discovery_url` is not available, I doubt get to 503! For example, in the first place, when we want to get the token to pass the authentication stage for auth-depended endpoints, we get the 404 when the `discovery_url` is not accessible! We don't get to the next level to get 503.

```bash
freva-client auth -u janedoe --host localhost:7777
Give password for server authentication: 
[2025-03-29T19:49:51] ERROR    freva-client: Failed to authenticate: Not Found                               
                      ERROR    freva-client: Token creation failed 
```
and on server logs:
```bash
2025-03-29T18:49 ERROR    freva-rest -                               auth.py:273
                          HTTPSConnectionPool(host='localhost',                 
                          port=8089): Max retries exceeded with url:            
                          /realms/Freva/.well-known/openid-configura            
                          tion (Caused by                                       
                          NewConnectionError('<urllib3.connection.HT            
                          TPSConnection object at 0xffff9bceddf0>:              
                          Failed to establish a new connection:                 
                          [Errno 111] Connection refused'))                     
                 INFO     uvicorn.access -                 httptools_impl.py:476
                          192.168.65.1:43020 - "POST                            
                          /api/freva-nextgen/auth/v2/token                      
                          HTTP/1.1" 404 
```

freva-rest/docker-scripts/entrypoint.sh
antarcticrainforest: Yes, I've fixed that.

antarcticrainforest:
There was a braking change update upstream in xarray which led xpublish to fail. In order to fix this I decided to drop xpublish as a dependency and do the zarr serialization directly here. 

In hindsight I could have used a different branch but it is what it is.

mo-dkrz (approved):
LGTM!

btw dropping xpublish was smart! I would love the day you tell me, you dropped the whole `zarr` dependencies from our infra! Then we can get rid of these monkey-patch of downgrading the zarr version to keep our infra alive...

PR #148, opened 2025-04-09
switch to the most recent zarr version

mo-dkrz:
This morning, while working on the STAC-API endpoint development, I encountered an issue after refreshing my environment. The zarr dependency began throwing errors and was not functioning properly, which ended up blocking my developement. To resolve this, I had to dig into the library and make some necessary fixes.

In this small PR, we’ve reverted from `zarr<3` back to the latest release of `zarr`, and all related adjustments have been made accordingly.

When I was doing this changes, I noticed that in `Zarr-3`, most of the core functions have been refactored to be asynchronous. So, if we plan to fully adopt Zarr-3 and use its generic functionality, the proper long-term solution would be to refactor our data loader worker functions to be async as well, allowing us to `await` Zarr's internal async methods properly.

But, I still strongly fan for maintaining our own `zarr_utils` abstraction instead of relying directly on Zarr’s generics funcs. The Zarr codebase tends to introduce breaking changes frequently and without much notice, which can easily disrupt our workflow and break core functionality.

### DIFF
--- a/freva-data-portal-worker/pyproject.toml
+++ b/freva-data-portal-worker/pyproject.toml
@@ -37,7 +37,7 @@ dependencies = [
 "rioxarray",
 "watchfiles",
 "xarray",
-"zarr<3",
+"zarr",
 ]
 [project.scripts]
 data-loader-worker = "data_portal_worker:cli.run_data_loader"
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,4 @@ dev-env/mongodb
 dev-env/certs
 *keycloakdb.lock.db
 *keycloakdb.*
+.DS_Store
--- a/freva-data-portal-worker/src/data_portal_worker/load_data.py
+++ b/freva-data-portal-worker/src/data_portal_worker/load_data.py
@@ -11,7 +11,7 @@ import redis
 import xarray as xr
 from dask.distributed import Client, LocalCluster
 from xarray.backends.zarr import encode_zarr_variable
-from zarr.storage import array_meta_key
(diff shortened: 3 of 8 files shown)


freva-data-portal-worker/pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
 "rioxarray",
 "watchfiles",
 "xarray",
-"zarr<3",
+"zarr",
```
antarcticrainforest, line 40: Will those changes be working with all zarr versions on >=3 ? Maybe we'll have to go to >=3 instead of the other way round?

mo-dkrz: true

antarcticrainforest (commented):
I am only concerned about the zarr version comparability. Is this working with zarr ==2.X? Because your dependency means that if we update the data-loader service, we will *not* update zarr to 3.X . Will it still work then?

Other than that LGTM!

freva-data-portal-worker/pyproject.toml
mo-dkrz: I think we cannot do it! Because the most recent version supports only [py3.11 and greater](https://github.com/zarr-developers/zarr-python/blob/main/pyproject.toml#L33) and if we change, our `<py3.11` fails.

mo-dkrz:
@antarcticrainforest Now it supports all; 2.X and 3.X

antarcticrainforest (commented):
LGTM!

PR #151, opened 2025-04-20
STAC-API based on Solr and compatible with Freva endpoints - BASE

mo-dkrz:
Since this feature involved a big volume of changes, I decided to split the implementation into smaller pull requests to reduce the burden on reviewer and on the other hand we can resolve the comments and merge it faster and more efficient. 

This PR lays the basis for the STAC-API feature integration. It establishes the base endpoints, which are now fully compatible with the existing Freva nextgen API structure. Additionally, the search functionality has been implemented using Solr. Most of the endpoints defined in the STAC API spec are included in this PR.

* Remaining Features (to be added in a follow-up branch rebased from this one):

- [x] `search` endpoint via GET method  
- [x] `search` endpoint via POST method  
- [x] `queryables` endpoint  
- [ ] Since each project is treated as a collection in this design, we’ll need to create a new Solr core to store constant metadata for each project (e.g., title, description). We may also include temporal and spatial extents, as well as thumbnails, to improve the catalog's presentation.  
- [ ] Optionally, we might explore how to support **STAC extensions**, although this is not a priority at this stage. (it's just a fancy addition) 
- [ ] Add more STAC assets in collections and dynamically assign accurate temporal and spatial metadata to each collection.

How to try it out:

This branch is already deployed on the sandbox machine. You can explore the endpoints directly via the STAC Browser:  https://freva-clint.github.io/stac-browser/#/external/freva-sandbox.cloud.dkrz.de/api/freva-nextgen/stacapi/?.language=en

And all available STAC-API endpoints are documented and accessible via the API helper:  
https://freva-sandbox.cloud.dkrz.de/api/freva-nextgen/help#tag/STAC-API


mo-dkrz:
Since people usually use the STAC API endpoints or web or SATC-Client or PySTAC for their research, to implement the freva CLI and python, I didn't have an appropriate idea; so I postponed it to find a proper way later on. If you also have any idea on designing the CLI and Py client please let me know.

freva-rest/src/freva_rest/stac_api/core.py
```diff
@@ -0,0 +1,750 @@
...
+
+        caller_name = sys._getframe(1).f_code.co_name
+
+        for key in query:
+            if (
+                key not in ["datetime", "bbox", "limit", "token"]
+            ) and caller_name == "collection_items":
+                raise HTTPException(status_code=422, detail="Could not validate input.")
+
+        return STACAPI(
```
antarcticrainforest, line 93: `cls(` ?

freva-rest/src/freva_rest/stac_api/core.py
```diff
@@ -0,0 +1,750 @@
...
+        datetime: Optional[str] = None,
+        bbox: Optional[str] = None,
+        uniuq_key: Literal["file", "uri"] = "file",
+        **query: list[str],
+    ) -> None:
+        self.config = config
+        self.uniq_key = uniuq_key
+        self.solr_object = Solr(config)
+        self.stac_object = STAC(config)
+        self.solr_query = self.solr_object.query
```
antarcticrainforest, line 41: Could you define search method in the solr core object that you could call here, instead of mingling with the query dict?

If we really want to use different search back ends you then this class shouldn't know about the search backend it is using.

antarcticrainforest (changes requested):
I have mainly a couple of questions.

One concern. The `stac_api.core` module directly uses the solr queries. I think this stat api code shouldn't have to worry about which search backend is in use. Meaning we'll properly have to define a couple of methods in the solr core that the stac core calls.

freva-rest/src/freva_rest/stac_api/endpoints.py
```diff
@@ -0,0 +1,210 @@
...
+from .core import STACAPI
+from .schema import CONFORMANCE_URLS, STACAPISchema
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
```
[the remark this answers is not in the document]
mo-dkrz: no worries about this. I've written in a comment above that it is only temp for the sake of demonstration and development. It's a problem with STAC-browser. The same issue happens with `localhost` on dev as well. when the header doesn't have the CORS, you get this error:
`This issue may occur when servers don't allow external access via web browsers (e.g., when CORS headers are not present). Please contact the service operator to resolve this issue. `

On my last commit it would be removed!

mo-dkrz:
I close this to not spam your email and open when it's ready

freva-rest/src/freva_rest/api.py
```diff
@@ -1,8 +1,15 @@
...
 from freva_rest.auth import *  # noqa: F401
+from freva_rest.config import ServerConfig
 from freva_rest.databrowser_api.endpoints import *  # noqa: F401
 from freva_rest.freva_data_portal.endpoints import *  # noqa: F401
 from freva_rest.rest import app
 
+from .logger import logger
+
+server_config = ServerConfig()
+if "stacapi" in server_config.services:
```
mo-dkrz, line 11: One might doesn't tend to list the STACAPI endpoints in their helper if they don't like to use this service. So that is why it has been made through if statement in `api.py`

tests/test_databrowser_cli.py
```diff
@@ -125,11 +125,13 @@ def test_stac_catalogue(
                 item_content = zip_file.read(member)
                 temp_item_path = temp_dir / "test_item.json"
                 temp_item_path.write_bytes(item_content)
-                subprocess.run(
+                res_stac = subprocess.run(
                     ["stac-check", str(temp_item_path)],
                     check=True,
                     capture_output=True,
                 )
+                assert "ITEM Passed: True" in res_stac.stdout.decode("utf-8")
```
mo-dkrz, line 133: we were getting broken test on STAC-CHECK since they changed the `ITEM Valid` to `ITEM valid`! resolved now.

mo-dkrz:
I think it seems like to be ready to go for the next round of review.
Since still I'm not totally satisfied with integrating this feature through iframe in the front-end, I didn't deploy it on sandbox yet. When the front end is ready I write you here. The reason I'm not happy, is that, we don't see the stacapi params and paths on the link bar. For example we go clicking through collections and items, but the link bar is always stays the same `https://<proxy>/stacbrowser` due to the iframe concept that is not able to pass the path to main link.

freva-rest/src/freva_rest/stac_api/schema.py
```diff
@@ -0,0 +1,132 @@
...
+class STACAPISchema:
+    """Class holding all apache solr config parameters."""
+
+    params: Dict[str, Any] = {
+        "limit": Query(
+            alias="max-results",
+            title="Max. results",
+            description="Control the number of maximum result items returned.",
+            ge=0,
+            le=1500,
```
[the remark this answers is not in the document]
mo-dkrz: it has changed to 1000, the standard of STACAPI

freva-rest/src/freva_rest/databrowser_api/__init__.py
```diff
@@ -2,7 +2,17 @@
 
 from freva_rest import __version__
 
-__all__ = ["__version__"]
+from .core import Solr
+from .stac import STAC, Asset, Item, Link
```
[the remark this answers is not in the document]
mo-dkrz: Apart from `from .core import Solr` all the rest has been dropped. To STACAPI be able to talk with the backend, there won't be any other way than keeping import Solr in init.

mo-dkrz:
@antarcticrainforest ready to merge!

antarcticrainforest:
LGTM

PR #156, opened 2025-05-08
Zarr streaming pagination for web

mo-dkrz:
In this PR we are going to add the zarr-streaming parameter in the `extended-search` endpoint for web purposes.
In the beginning I decided to paginate the `load` endpoint to have max-result on each params, to be able to see the zarr steam `uri`s on the file container of databrowser. But while it was totally feasible, then I found, adding pagination on an streaming endpoint, won't be a good idea since we don't use the generic param of the search engine and we can only stop the iteration via designing a counter. I didn't feel good about that approach. So I went through adding the `zarr-stream` param in `extended-search`
To achieve the final goal on web, another important thing has been changed! `required_dependency`: Since `extended-search` now works with(for zarr-streaming) and without authentication, we needed an optional auth instance creation. So i went through adding optional argument to this function.

Now one off-topic idea: when I was implementing this, another thing that crossed to my mind, it's concatenating the `data-search` with `load`, `intake-catalog` and `stac-catalogue`. With this idea, we can simply add a `zarr-stream=True` and `catalogue-type=intake/stac` parameters in data search endpoint and enable and cover the whole functionality of the `load` , `intake-catalog` and `stac-catalogue` endpoints in data-search which makes our codebase way leaner. 
What do you think about this idea?

### DIFF
--- a/freva-rest/src/freva_rest/auth.py
+++ b/freva-rest/src/freva_rest/auth.py
@@ -85,8 +85,8 @@ class SafeAuth:
             if self._auth is None and await self._check_server_available():
                 self._auth = Auth(self.discovery_url)
 
-    def required_dependency(
-        self,
+    def create_auth_dependency(
+        self, required: bool = True
     ) -> Callable[
         [SecurityScopes, Optional[HTTPAuthorizationCredentials]],
         Awaitable[IDToken],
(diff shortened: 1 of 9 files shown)


freva-rest/src/freva_rest/auth.py
```diff
@@ -103,22 +103,34 @@ def required_dependency(
...
         async def dependency(
             security_scopes: SecurityScopes,
             authorization_credentials: Optional[
                 HTTPAuthorizationCredentials
-            ] = Depends(HTTPBearer()),
+            ] = Depends(HTTPBearer(auto_error=required)),
         ) -> IDToken:
             await self._ensure_auth_initialized()
 
             if self._auth is None:
```
antarcticrainforest, line 114: wouldn't it be enough to simply to a `if self._auth is None and required:` ? And the rest stays the same?

freva-rest/src/freva_rest/databrowser_api/core.py
```diff
@@ -980,28 +981,47 @@ async def extended_search(
...
+                    )
+                    doc[self.uniq_key] = f"{api_path}/{uuid5}.zarr"
+                except Exception as pub_err:
+                    logger.error(
+                        "Failed to publish to Redis for %s: %s", uri, pub_err
+                    )
+                    doc[self.uniq_key] = (
+                        "Internal error, service not able to publish"
+                    )
+                doc["fs_type"] = doc.get("fs_type", "posix")
```
antarcticrainforest, line 1018: How about using an external func for this? Then you can use this external func here and in the data-portal endpoint.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -291,10 +291,15 @@ async def extended_search(
     multi_version: Annotated[bool, SolrSchema.params["multi_version"]] = False,
     translate: Annotated[bool, SolrSchema.params["translate"]] = True,
     max_results: Annotated[int, SolrSchema.params["batch_size"]] = 150,
+    zarr_stream: bool = False,
```
antarcticrainforest, line 294: Follow annotation the logic from the rest.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -304,8 +309,17 @@ async def extended_search(
         translate=translate,
         **SolrSchema.process_parameters(request),
     )
+    if (
+        zarr_stream
+        and current_user is None
+        and "zarr-stream" in server_config.services
```
antarcticrainforest, line 315: Have you thought about the case where `zarr-stream` is requested but it's not supported by the server, meaning that it's not in the `services` flag. This flag is for instances that don't want to go through the pain of setting up an extra redis and streaming worker. I assume that FUB woulnd't be interested in this feature so I don't want to make them deploy stuff they don't need.

I would do it like this:

```python
if zarr-stream not in server_config:
   zarr_stream = False
   
 if current_user is None and zarr_stream:
   raise HTTPException(
                        status_code=401,
                        detail="Zarr-streaming needs authentication",
                    )
```

antarcticrainforest (changes requested):

freva-rest/src/freva_rest/auth.py
mo-dkrz: i think it doesn't work to simply add the `required` in `if self._auth:`:

1.  `auto_error=required` arg from `HTTPBearer` is necessary to handle the 401 for cases that we don't need bearer token since by default it's True (which means by default we need to provide the bearer token otherwise it raises 401 - or better said refuse to xreate the auth- even for cases that creation of auth instance is not required!).

2. the same story for try except statement in `self._auth.required(security_scopes, authorization_credentials)`

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -309,10 +314,11 @@ async def extended_search(
...
         **SolrSchema.process_parameters(request),
     )
+    if "zarr-stream" not in server_config.services:
+        zarr_stream = False
     if (
         zarr_stream
         and current_user is None
-        and "zarr-stream" in server_config.services
     ):
         return JSONResponse(
```
antarcticrainforest, line 323: Why don't you want to raise an HttpException here?

antarcticrainforest (changes requested):

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -320,10 +320,12 @@ async def extended_search(
         zarr_stream
         and current_user is None
     ):
-        return JSONResponse(
-            content={"detail": "Not authenticated"},
-            status_code=status.HTTP_401_UNAUTHORIZED
+        logger.exception("User not authenticated for zarr stream.")
```
antarcticrainforest, line 323: Nit-Pick: This can be an error or warning, we don't really need a traceback. Other than that. LGTM

antarcticrainforest (approved):

mo-dkrz:
@antarcticrainforest shall I release since you might need the ping endpoint on deployment?

PR #180, opened 2025-06-28
Make auth methods more convenient.

antarcticrainforest:
I noticed that when users want to use the OAuth2 token that was previously created  they have (according) to the docs use the `authenticate` method. This is for example done when they want to access zarr stream data in xarray.

But there is a caveat: the `authenticate` method will always spawn a browser based auth session, regardless of whether or not the token is still valid.

To stop that I instructed `authenticate` to use existing tokens, if possible. I've also created a convenience property of the databrowser class to directly access to OAuth2 token.

### DIFF
--- a/freva-client/src/freva_client/query.py
+++ b/freva-client/src/freva_client/query.py
@@ -28,6 +28,7 @@ from rich import print as pprint
 
 from .auth import Auth
 from .utils import logger
+from .utils.auth_utils import Token, choose_token_strategy, load_token
 from .utils.databrowser_utils import Config, UserDataHandler
 
 __all__ = ["databrowser"]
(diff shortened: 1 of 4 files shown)


freva-client/src/freva_client/query.py
```diff
@@ -385,6 +388,42 @@ def _create_intake_catalogue_file(self, filename: str) -> None:
...
+        .. code-block:: python
+
+            from freva_client import databrowser
+            import xarray as xr
+            db = databrowser(dataset="cmip6-hsm", stream_zarr=True)
+            dset = xr.open_dataset(
+               zarr_files[0],
+               chunks="auto",
+               engine="zarr",
+               storage_options={"header":
```
mo-dkrz, line 413: `headers` instead of `header`?

mo-dkrz (changes requested):
minor comments:
- in the `databrowser_cli`, I think we have the same `-t` flag for both `time` and `token`. One of them should have to be changed or removed.
other than this and the following comments, LGTM!

freva-client/src/freva_client/query.py
antarcticrainforest: yes!!!

antarcticrainforest:
Thanks, I think I should have addressed all comments.

antarcticrainforest:
Is this ok now?

mo-dkrz (approved):
LGTM!

PR #184, opened 2025-07-04
Quick fix

antarcticrainforest:
This fixes some k8s deployment issues (hopefully) I'd need to check first so it doesn't need to be merged straight away.


mo-dkrz (approved):
so far LGTM!

PR #191, opened 2025-07-18
Custom flavour

mo-dkrz:
Add support for custom flavour mappings via

- Env var `API_TRANSLATOR_MAPPING_<NAME>=<JSON>`
- `api_config.toml` under `[restAPI.flavour_mappings.<name>]`
- Remove built‑in `nextgems` mapping (has to be enabled explicitly)

```env
API_TRANSLATOR_MAPPING_KLIMAKATASTER={"project": "projekt","product": "produkt","institute": "institut","model": "modell","experiment": "experiment","time_frequency": "zeit_frequenz","realm": "bereich","variable": "variable","ensemble": "ensemble","time_aggregation": "zeit_aggregation","fs_type": "fs_typ","grid_label": "gitter_label","cmor_table": "cmor_tabelle","driving_model": "antriebsmodell","format": "format","grid_id": "gitter_id","level_type": "level_typ","rcm_name": "rcm_name","rcm_version": "rcm_version","dataset": "datensatz","time": "zeit","bbox": "bbox","user": "benutzer"}
```
or 

```toml
[restAPI]
services = ["databrowser", "zarr-stream", "stacapi"]
proxy = ""
[restAPI.flavour_mappings.nextgems]
experiment = "simulation_id"
ensemble = "member_id"
fs_type = "fs_type"
grid_label = "grid_label"
institute = "institution_id"
model = "source_id"
project = "project"
product = "experiment_id"
realm = "realm"
variable = "variable_id"
time = "time"
bbox = "bbox"
time_aggregation = "time_reduction"
time_frequency = "time_frequency"
cmor_table = "cmor_table"
dataset = "dataset"
driving_model = "driving_model"
format = "format"
grid_id = "grid_id"
level_type = "level_type"
rcm_name = "rcm_name"
rcm_version = "rcm_version"
```

this branch has been already deployed on sandbox. https://freva-sandbox.cloud.dkrz.de/databrowser/?flavour=klimakataster


antarcticrainforest (changes requested):
This PR basically breaks the whole idea of the server client architecture. For the record: If the client needs to communicate settings with the server, it should do so via rest api endpoints.

Also I am not a fan of the current outline. How about you create dedicated endpoints that `add` and `query` flavors. Then users can add their own flavour definitions. 

When a flavour gets added to the restAPI we store that flavour with the username to the MongoDB. To make flavours applicable for all users we have a custom key like `general` or `all` or something instead of the user id. General flavours can only be added by either pre-defined admin users or tokens that have a pre-defined token admin claims.

In terms of client. You want to add a function / cli option to add flavours. Once added via the endpoint the new flavours are added to the freva.toml config file. Whenever the client runs, it queries the available flavours and checks if the config file needs updating. That way you can display flavours in the cli options without querying the server (you don't know the sever yet). But of course feel free to think of a better solution.

freva-rest/src/freva_rest/config.py
```diff
@@ -440,6 +397,33 @@ def mongo_collection_userdata(self) -> AsyncIOMotorCollection:
...
+            return True
+
+        claims = current_user.dict()
+        roles = (
+            claims
+            .get("resource_access", {})
+            .get("realm-management", {})
+            .get("roles", [])
+        )
+        return "realm-admin" in roles or "admin" in roles
```
antarcticrainforest, line 425: Either you make this around a predefined list of admins users and check if the current token user name in the admin list *or* you make this based on token claims. Then instead to setting admin user names you set a regex for the token claims and match it here. The advantage is, that once set you can go to keycloak and add users there. No restart, no redeployment necessary. Hence I'd prefer that method.

freva-rest/src/freva_rest/databrowser_api/core.py
```diff
@@ -625,19 +665,18 @@ async def validate_parameters(
...
-                and key not in ("time_select", "bbox_select", "zarr_stream")
-                + cls.uniq_keys
-            ):
-                raise HTTPException(status_code=422, detail="Could not validate input.")
+        user_name = current_user.preferred_username if current_user else "global"
+
+        translator = await cls._validate_and_get_flavour(config, flavour, user_name)
+        translator.translate = translate
+
+        valid_facets = translator.valid_facets + (["version"] if multi_version else [])
```
antarcticrainforest, line 677: So version always stays version? Do you ensure that in the translation?

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -83,20 +57,24 @@ async def overview() -> SearchFlavours:
...
+    for flavour_response in all_flavour_responses:
+        flavour_name = flavour_response.flavour_name
+        if flavour_name in ("cordex",):
+            attributes[flavour_name] = list(flavour_response.mapping.values())
         else:
-            attributes[flavour] = [
-                f
-                for f in translator.forward_lookup.values()
-                if f not in translator.cordex_keys
+            cordex_keys = ("rcm_name", "driving_model", "rcm_version")
```
antarcticrainforest, line 72: Don't do this. 

Don't start with if something is something very specific then to this.

Abstract, and find a solution with code that's not checking for specific edge cases.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -156,12 +138,15 @@ async def metadata_search(
     response_class=PlainTextResponse,
 )
 async def data_search(
-    flavour: str = Depends(validate_flavour_dependency),
-    uniq_key: Literal["file", "uri"] = Path(..., description="core type"),
```
antarcticrainforest: Why did you delete the Path?

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -511,3 +505,158 @@ async def delete_user_data(
...
+        example="nextgem"
+    ),
+    owner: Optional[str] = Query(
+        None,
+        description="Filter by owner ('global' or username)",
+        example="global"
+    ),
+    request: Request = Required,
+    current_user: Optional[TokenPayload] = Security(
+        auth.create_auth_dependency(required=False), scopes=["oidc.claims"]
```
antarcticrainforest, line 583: I am not sure why we need a claim check here.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -255,13 +245,16 @@ async def intake_catalogue(
...
-    uniq_key: Literal["file", "uri"] = Path(..., description="core type"),
+    flavour: str,
+    uniq_key: Literal["file", "uri"],
     start: Annotated[int, SolrSchema.params["start"]] = 0,
     multi_version: Annotated[bool, SolrSchema.params["multi_version"]] = False,
     translate: Annotated[bool, SolrSchema.params["translate"]] = True,
     max_results: Annotated[int, SolrSchema.params["max_results"]] = -1,
     request: Request = Required,
+    current_user: Optional[TokenPayload] = Security(
+        auth.create_auth_dependency(required=False), scopes=["oidc.claims"]
```
antarcticrainforest, line 256: I am not sure why we need a claim check here.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -201,13 +187,16 @@ async def data_search(
...
```
antarcticrainforest, line 198: I am not sure why we need a claim check here.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -156,12 +138,15 @@ async def metadata_search(
...
-    flavour: str = Depends(validate_flavour_dependency),
-    uniq_key: Literal["file", "uri"] = Path(..., description="core type"),
+    flavour: str,
+    uniq_key: Literal["file", "uri"],
     start: Annotated[int, SolrSchema.params["start"]] = 0,
     multi_version: Annotated[bool, SolrSchema.params["multi_version"]] = False,
     translate: Annotated[bool, SolrSchema.params["translate"]] = True,
     request: Request = Required,
+    current_user: Optional[TokenPayload] = Security(
+        auth.create_auth_dependency(required=False), scopes=["oidc.claims"]
```
antarcticrainforest, line 148: I am not sure why we need a claim check here.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -110,12 +88,15 @@ async def overview() -> SearchFlavours:
...
-    flavour: str = Depends(validate_flavour_dependency),
-    uniq_key: Literal["file", "uri"] = Path(..., description="core type"),
+    flavour: str,
+    uniq_key: Literal["file", "uri"],
     multi_version: Annotated[bool, SolrSchema.params["multi_version"]] = False,
     translate: Annotated[bool, SolrSchema.params["translate"]] = True,
     facets: Annotated[Union[List[str], None], SolrSchema.params["facets"]] = None,
     request: Request = Required,
+    current_user: Optional[TokenPayload] = Security(
+        auth.create_auth_dependency(required=False), scopes=["oidc.claims"]
```
antarcticrainforest, line 98: I am not sure why we need a claim check here.

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -511,3 +505,158 @@ async def delete_user_data(
...
+            "Attempt to delete built-in flavour: %s by user: %s",
+            flavour_name,
+            current_user.preferred_username
+        )
+        raise HTTPException(
+            status_code=422,
+            detail=f"Cannot delete built-in flavour '{flavour_name}'"
+        )
+
+    is_global = server_config.is_admin_user(current_user)
```
antarcticrainforest, line 655: This assumes that admins can only add global falvours. Why?

antarcticrainforest: There is a grammatically nuance in my comment. I said:

Admins can *only* delete global flavours.

I guess you understood:

*Only* admins can delete global flavours.

I am not questioning at all what you said. But what I said was the following: When I am admin my flavour will always become global. This logic could be ok but then you'd have to also implement this logic in the flavour add functionality. Because the add functionality checks if you're admin *and* your flavour should become global. Sot there is a chance for an admin to add a private flavour. Here we only check if you're admin and assume if we are then it's global. So there is a clash in the code.

mo-dkrz:
@antarcticrainforest Thanks for the review.
I’ve been thinking about your idea regarding the default flavour.
The idea is that when a project deploys Freva, it may want to use a custom global flavour instead of the default `"freva"`.
Examples:
`"klimakataster"` flavour for the RegIKlim project
`"nextgems"` flavour for the NextGEMS project

This means that in all clients (cli, py, and Web), the project’s default flavour should be shown consistently.

there are two ways to achieve this:

1. client-side; web-only

we define the default flavour only in `freva-web` config env and users would see the project’s default flavour in the web interface. but the limitation is that, `freva-client` would still show the general `"freva"` flavour and not respect the project’s default. There is no possible way on cli and python

2. server-side

we define a `default_flavour `variable in the config. But since flavour is currently a **path parameter** in all endpoints, we would need to:
a) Refactor endpoints to remove flavour from the path,
b) Add flavour as a query parameter instead.
Then through this way, this makes handling the project’s default flavour more straightforward.
but the main concern is backward compatibility! For a transition period, we should support both the current path parameter and the new query parameter. and alos a deprecation warnings should be added for user who are still using the path param, and at the end we can drop flavour from the path after a while or dedicated due time, once clients have migrated.


Which way do you prefer to keep up with?

antarcticrainforest:
> > We should add and document a functionality to read the default flavour (if set) from the config.
> 
> Actually, here is the part that’s not 100% clear for me. On the **web client**, there’s no real problem — the project admin sets the default flavour once and it just works. The main issue is with the `freva-client`. Even if we document it, we can’t really expect new users to read through the docs just to understand this detail.
> 
> Simple story: I’m a new RegIKlim user who just installed `freva-client` and I want to search RegIKlim data. I do:
> 
> ```shell
> pip install freva-client
> freva-client databrowser metadata-search Zeitfrequenz="1h" Zeit="1970-12-20%20TO%202025-12-20" --host https://www-regiklim.dkrz.de
> ```
> 
> From user perspective, this should just work! it’s the same as what user sees in the web databrowser command window. But instead, I get a 422 error, because the search keys don’t match the default `"freva"` flavour.
> 
> At this point, user needs to contact the project admin, who then points user to the docs, and only then I can fix it.
> 
> If we add this functionality on the server-side, then whenever the user points to a project host (like RegIKlim), the web service would automatically know which default flavour to use. It’s a bit more work for us in code, but it makes things much simpler and smoother for the user, which is our final goal in Freva.
> 
> based on this story, do you still believe in the changing on client? if so, let's go to change on client.

Those metadata flavours are supposed to be requested from client side. This is where they should stay. 

I could turn your argument about by saying that having the default flavours defined server side and changing them will result in the screw up of workflows of basically *all* users. 

Also, in your argument you've forgotten the default case: users use modules to load the software. Here the admin can define what's supposed to be default behaviour and what not. All other approaches, such as users installing the client lib should be considered as advanced. We can only provide a limited amount to support in such cases. This includes how to configure the system properly.

PR #196, opened 2025-07-30
Fix userdata leaking issue and manage the STAC Item-asset better

mo-dkrz:
In this PR we are going to fix a minor issue on user-data and also we try to make the help section on local access asset of Items on both STAC API and STATIC better.


antarcticrainforest:
Pleas don't forget to bump the version as we need pypi updates.

PR #198, opened 2025-08-07
fix(docs): link to freva legacy

eelucio:
currently (since the project move and freva repo renaming) there is no easy way to access old freva documentation as the only direct reference is this one link.


antarcticrainforest (changes requested):
Thanks, good catch.

Can you please also change the intro to the auth docs?
https://freva-nextgen.readthedocs.io/en/latest/auth/auth-lib.html#authentication-using-the-freva-client-library it's completely out of place.

PR #199, opened 2025-08-08
auth: normalise env databrowser.host url to https with no port

eelucio:
this little function normalises the default databrowser.host so it always ends at https, it also removes the port (7777) number that would not be shown in https

### DIFF
--- a/freva-client/src/freva_client/auth.py
+++ b/freva-client/src/freva_client/auth.py
@@ -294,7 +294,7 @@ def authenticate(
 
     Parameters
     ----------
-    refresh_token: str, optional
+    token_file: str, optional
         Instead of setting a password, you can set a refresh token to refresh
         the access token. This is recommended for non-interactive environments.
     host: str, optional
@@ -308,15 +308,11 @@ def authenticate(
 
     Examples
     --------
-    Interactive authentication:
+    Currently, a new token can only be created *from scratch* via the website of
+    your host instance. Once logged in, click the **Token button** to generate a
+    refresh token, then copy, paste, or download it.
 
-    .. code-block:: python
-
-        from freva_client import authenticate
-        token = authenticate()
-        print(token)
-
-    Batch mode authentication with a refresh token:
+    Authentication with a refresh token:
 
     .. code-block:: python
 
--- a/freva-client/src/freva_client/auth.py
+++ b/freva-client/src/freva_client/auth.py
@@ -308,11 +308,15 @@ def authenticate(
 
     Examples
     --------
-    Currently, a new token can only be created *from scratch* via the website of
-    your host instance. Once logged in, click the **Token button** to generate a
-    refresh token, then copy, paste, or download it.
+    Interactive authentication:
 
-    Authentication with a refresh token:
+    .. code-block:: python
+
+        from freva_client import authenticate
+        token = authenticate()
+        print(token)
+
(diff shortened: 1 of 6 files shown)


freva-client/src/freva_client/auth.py
```diff
@@ -246,30 +246,38 @@ def authenticate(
...
+        def _https_no_port(url: str) -> str:
+            """Normalize only when no explicit host was passed."""
+            p = urllib.parse.urlsplit(url)
+            netloc = p.hostname or ""
+            return urllib.parse.urlunsplit(
+                ("https", netloc, p.path, p.query, p.fragment)
+            )
+
         cfg = config or Config(host)
+        use_auth_url = cfg.auth_url if host else _https_no_port(cfg.auth_url)
```
antarcticrainforest, line 259: The config class should take care of this. Could you quickly explain why you decided to do that?

antarcticrainforest (changes requested):
I am a but worried that the auth changes you made will break the logic of the config class.

freva-client/src/freva_client/auth.py
eelucio: On my tests I had issues when the auth was done via http hostname  and not https. 

It could login but it would never be able to retrieve the token, not in levante, not locally (I tested exporting a evaulation.conf file while in the VPN).

This takes the url given by the config class (databrowser.host or solr.host) and replaces http by https and removes the port. e.g.

`http://hostname:7777--> https://hostname`

antarcticrainforest: But your changes assume that the rest-api is *always* to be reached by a reverse proxy. Which is not the case. The solution to the issue is not assigning a default port of 7777 to the auth_url.

antarcticrainforest: I've just checked the code and it doesn't assume any default port:

https://github.com/freva-org/freva-nextgen/blob/c222057ecce5a358f3c319d2f055c05eefe4c2dc/freva-client/src/freva_client/utils/databrowser_utils.py#L50

So the issue can be solved by configuring the correct databrowser host/url

eelucio: as far as I understand this lines
https://github.com/freva-org/freva-nextgen/blob/c222057ecce5a358f3c319d2f055c05eefe4c2dc/freva-client/src/freva_client/utils/databrowser_utils.py#L55-L57

will prioritise databrowser.host over sor.host
databrowser.host afaik is indicated with hostname:port (at least in most of our conf files)

in any case the function  it just does 2 things: 1) replace http by https and 2) ditches the port if exists.

```
In [1]: import urllib

In [2]:         def _https_no_port(url: str) -> str:
   ...:             """Normalize only when no explicit host was passed."""
   ...:             p = urllib.parse.urlsplit(url)
   ...:             netloc = p.hostname or ""
   ...:             return urllib.parse.urlunsplit(
   ...:                 ("https", netloc, p.path, p.query, p.fragment)
   ...:             )
   ...:

In [7]: url = "http://hostname.com:7777"

In [8]: _https_no_port(url)
Out[8]: 'https://hostname.com'

In [9]: url = "http://hostname.com"

In [10]: _https_no_port(url)
Out[10]: 'https://hostname.com'
```

but as I said, this did not solve the issue with e.g. regiklim.

PR #201, opened 2025-08-15
Give users the choice to set the login timeout

antarcticrainforest:
With this users can choose the login time. When port forwarding is needed or other things like token challenges are involved in the login process 30 secs of timeout are too little. With this users can increase the timeout or even set it to wait forever.

On top of the information displayed on who to do port forwarding is improved.

Thanks for quick review @mo-dkrz


mo-dkrz (approved):
LGTM!

PR #205, opened 2025-08-29
Adding default flavour for client

mo-dkrz:
Here, we are going to introduce the default flavour option on the client side config file (adding `default_flavour` in `freva.toml` under the `freva` table; as same as the `host`) and also there were bunch of minor issues that all are fixed as well.

### DIFF
--- a/freva-client/src/freva_client/utils/auth_utils.py
+++ b/freva-client/src/freva_client/utils/auth_utils.py
@@ -8,6 +8,7 @@ import time
 from pathlib import Path
 from typing import Literal, Optional, TypedDict, Union, cast
 
+import requests
 from appdirs import user_cache_dir
 
 TOKEN_EXPIRY_BUFFER = 60  # seconds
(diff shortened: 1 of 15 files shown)


freva-client/src/freva_client/utils/auth_utils.py
```diff
@@ -238,3 +239,40 @@ def wait_for_port(host: str, port: int, timeout: float = 5.0) -> None:
...
+        Whether the request is for zarr data.
+    databrowser_url : str or None
+        The URL of the databrowser to query for available flavours.
+        If None, the function will skip querying and assume authentication
+        is required for non-default flavours.
+    """
+    if zarr:
+        return True
+    if not flavour or flavour in {"freva", "cmip6", "cmip5", "cordex", "user"}:
+        return False
```
antarcticrainforest, line 265:
```suggestion
    if flavour in {"freva", "cmip6", "cmip5", "cordex", "user", None}:
        return False
```

freva-client/src/freva_client/utils/auth_utils.py
```diff
@@ -238,3 +239,40 @@ def wait_for_port(host: str, port: int, timeout: float = 5.0) -> None:
...
+    try:
+        response = requests.get(f"{databrowser_url}/flavours", timeout=30)
+        response.raise_for_status()
+        result = {"flavours": response.json().get("flavours", [])}
+        if result and "flavours" in result:
```
antarcticrainforest, line 270:
```suggestion
        if "flavours" in result:
```

freva-rest/src/freva_rest/databrowser_api/endpoints.py
```diff
@@ -512,7 +513,7 @@ async def delete_user_data(
 
 @app.post(
     "/api/freva-nextgen/databrowser/flavours",
-    tags=["Data search"],
+    tags=["Flavour management"],
```
antarcticrainforest, line 516: That still belongs to data search.

antarcticrainforest (changes requested):

PR #207, opened 2025-08-30
Add well-known-endpoint.

antarcticrainforest:
For third party auth that has no information about the URL of the OIDC server it might be convenient to have the option to query the metadata of the OIDC server via the rest-api.

This PR adds the .well-known endpoint to do so.


mo-dkrz (approved):
LGTM!

PR #213, opened 2025-09-08
Switch to device based auth flow.

antarcticrainforest:
This implements the suggestion of using device based auth flow. Which is much more convenient for the users.


mo-dkrz (approved):
LGTM!

PR #215, opened 2025-09-10
Use header for client secrets.

antarcticrainforest:
@mo-dkrz this would be my suggestions for the hemlholtz AAI. We need to discuss what else we will need to do.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,6 +60,7 @@ session required pam_sss.so\n\
         PKGNAME=freva-rest && \
         cp /tmp/app/${PKGNAME}/src/freva_rest/api_config.toml /opt/${CMD}/config && \
         micromamba install -y -q -c conda-forge --override-channels \
+        'python<3.14' \
         aiohttp \
         cloudpickle \
        "cryptography=45.*" \


mo-dkrz:
Two quick questions for the latest implementation:

1. Client side question: 

**Current situation**: Different OIDC providers have conflicting requirements for the `prompt` parameter:
 i.Helmholtz AAI: Requires `prompt=consent` (does not accept `none`)
 ii. DKRZ: If we generalize and keep `prompt=consent` for all login cases, redirects back to login page even when SSO session is still valid, since it didn't define the consent page.

**problem**: This situation is being generalized and even for the DKRZ OIDC instance that doesn't need consent, client behaves the same, an it redirect the user to the login page.

**What we can do to keep it adjustable for different OIDC instances**: Moving `prompt` to a configuration variable in freva-rest config file/env var with `none` as default. Or since this is a client topic, we can put `prompt` setting in the `freva-client` config file, but I think this won't be a good idea to put the handling of such a basic security matter on the user burden.
Also we can keep current hardcoded behavior if this flexibility isn't required.

2. On RestfulAPI side: 

**Current situation:** the login endpoint hardcodes scopes with only `offline_access` as changeable. 

**Problem:** For example, I got a test on the web as well, and Helmholtz AAI requires explicit `email` scope to return email addresses. Otherwise it returns null. So it means, when it doesn't return the email, MySQL complains, due to the freva-legacy email-required design. We can simply fix this on the web-client, but this might happen for other users who probably want to use our rest-api to build a cli/web app that their OIDC instance uses different scope naming.

**What we can do:** we can accept `scope` as a query parameter in `/api/freva-nextgen/auth/v2/login` or as a surgical fix simply provide `email` in the current default `openid profile` along the others)
but if we add a `scope` instead of Boolean `offline_access` param, to the `login` endpoint, it makes the login API more flexible for custom CLI/web clients integrating with various OIDC providers that have different scope requirements.

PR #219, opened 2025-09-18
fix: config file precedence issue causing malformed host URLs

mo-dkrz:
When we were adding default_flavour support in client, the config reading logic incorrectly prioritized TOML files over INI files. This caused the system to use empty/invalid host values from default TOML configs (`host = "http://"`) instead of valid hosts from user INI configs (`databrowser.host = "https://actual-server.com"`).

The fix changes the config reading to collect host and flavour separately from multiple files, ensuring `Host` values come from INI files first (preserving existing user configs), and `Flavour` values can still be read from TOML files.

@antarcticrainforest could you please take a look at this. It needs to be released as well, since we need to update the pypi and conda versions. Shall i change the `freva-rest` version to 2509.2.0 as well?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 All notable changes to this project will be documented in this file.
+##[v2509.2.0]
+### Changed
+- Fix config file precedence issue causing malformed host URLs in freva-client
+
 ##[v2509.1.0]
 ### Changed
 - Switched to device login flow for python-lib and cli clients.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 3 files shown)


mo-dkrz:
Thanks @antarcticrainforest for the review.
The configuration file precedence logic has been restored to its initial design:
1. user config TOML in `~/.config/freva/freva.toml`
2. system config TOML in `env/share/freva/freva.toml` 
3. custom config via `FREVA_CONFIG` environment variable
4. eval system config from `EVALUATION_SYSTEM_CONFIG_FILE` environment variable

mo-dkrz:
It seems like, cryptography, 3 days ago has been released a version that is not compatible with existing [cffi](https://github.com/pyca/cryptography/blob/46.0.0/pyproject.toml#L8-L9) version in micromamba image of python3.13. I just downgraded and freeze the cryptography version for the time-being and then when they release an stable version we switch to the most recent version again.

PR #229, opened 2025-10-05
Add HTML format support to .zmetadata endpoint

mo-dkrz:
This PR adds a `format` parameter to the `.zmetadata` endpoint, allowing it to return either JSON (default) or Xarray's native HTML format.

Now that the web client can authenticate directly with freva-rest, we want to replace the metadata inspector functionality with zarr streaming. This requires displaying dataset metadata in the same HTML format that Xarray produces.
In the beginning we considered generating the HTML layout web-client-side from the JSON response, but this has drawbacks:
1. Any style changes in Xarray would require manual updates to maintain identical output
2. Reimplementing Xarray's HTML rendering would significantly increase client codebase complexity

By serving Xarray's HTML directly from the backend, we avoid duplicating styling logic and automatically stay in sync with Xarray's representation.

### DIFF
--- a/freva-rest/src/freva_rest/freva_data_portal/endpoints.py
+++ b/freva-rest/src/freva_rest/freva_data_portal/endpoints.py
@@ -2,12 +2,12 @@
 
 import asyncio
 import json
-from typing import Annotated, Any, Dict, List, Optional
+from typing import Annotated, Any, Dict, List, Optional, Union
 
 import cloudpickle
 from fastapi import Path, Query, Security, status
 from fastapi.exceptions import HTTPException
-from fastapi.responses import JSONResponse, Response
+from fastapi.responses import HTMLResponse, JSONResponse, Response
 from fastapi_third_party_auth import IDToken as TokenPayload
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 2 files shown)


freva-rest/src/freva_rest/freva_data_portal/endpoints.py
```diff
@@ -206,6 +207,13 @@ async def zemtadata(
             ),
         ),
     ],
+    format: Annotated[
```
antarcticrainforest, line 210: Are you sure that this parameter doesn't clash and will not clash in the future the zarr format?

antarcticrainforest (commented):
We do we need to do this here? I do understand that it might be convenient but I also feel like that this is not the right place here.

freva-rest/src/freva_rest/freva_data_portal/endpoints.py
mo-dkrz: @antarcticrainforest  Thanks for the comment. Yes, we can define a new endpoint to represent the HTML, which would help avoid potential issues in the future and benefit users accessing the endpoint directly via Jupyter Notebook or the web.

But, if we implement it entirely on the front end based on the `.zmetadata` endpoint’s JSON output and aim to match the xarray HTML format, we would need to rewrite the entire xarray [HTML formatting](https://github.com/pydata/xarray/blob/main/xarray/core/formatting_html.py) and [styling](https://github.com/pydata/xarray/tree/main/xarray) code. This would also require us to constantly update our implementation whenever xarray changes its codebase.

Additionally, `.zmetadata` does not provide enough information to fully replicate xarray’s output (e.g., coordinate values and other details). please let me know which approach you prefer?

PR #244, opened 2025-10-27
feat: logout option to the CLI and Python authorization

tropicrainforest:
Following up on the discussion in [the server-side logout endpoint PR](https://github.com/freva-org/freva-nextgen/pull/242#issuecomment-3442141999), this PR adds support for cleaning up both the local token file and the corresponding session on the OIDC server.
This was ready by the end of last week, I just forgot to push it earlier.

Main changes:

* we replaced

  ```bash
  freva-client auth
  ```

  with

  ```bash
  freva-client auth login
  ```
* And added

  ```bash
  freva-client auth logout
  ```

Since the latest REST API is already deployed on gems, you can test it with:

```bash
pip install git+https://github.com/freva-org/freva-nextgen.git@logout-client#subdirectory=freva-client
freva-client auth login --host https://www.gems.dkrz.de
freva-client auth logout --host https://www.gems.dkrz.de
```

Or in Python:

```python
from freva_client import authenticate, logout
authenticate(host="https://www.gems.dkrz.de")
logout(host="https://www.gems.dkrz.de")
```

### DIFF
--- a/docs/source/auth/auth-lib.rst
+++ b/docs/source/auth/auth-lib.rst
@@ -8,13 +8,14 @@ with the authentication system.
 
 .. autofunction:: freva_client.authenticate
 
+.. autofunction:: freva_client.logout
+
 .. _auth_cli:
 
 Using the command line interface
 ================================
 
-Token creation and refreshing can also be achieved with help of the ``auth``
-sub command of the command line interface
+Token creation and logout can be achieved with the ``auth`` sub command:
 
 .. code:: console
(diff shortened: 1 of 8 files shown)


antarcticrainforest:
I didn't mean to tell you, we don't need that. But only that I feel like the importance and complexity of this doesn't deserve a sub command by it's own. a `--revoke` flag (or parameter for that matter) would be enough in my opinion.

PR #318, opened 2026-03-14
Fix: when it begins, if key doesn't exist yet, job is queued (3), not gone (5)

tropicrainforest:
When `/html` is fired, the very first `/status` poll almost always wins the race against the worker's initial Redis write. The key doesn't exist yet, so the cache returns empty, and the backend defaults to `5` (gone). when in reality the job was just submitted. This returned `5` disappears within a second once the worker writes its first entry. On a small change, this PR is going to return 3 status, when the `Cache.get(key)` is still empty. Do you think if this makes sense?

### DIFF
--- a/freva-rest/src/freva_rest/freva_data_portal/utils.py
+++ b/freva-rest/src/freva_rest/freva_data_portal/utils.py
@@ -166,8 +166,10 @@ async def read_redis_data(
         timeout += 1
     npolls = 0.0
     dt = 0.5
-    data = cloudpickle.loads((await Cache.get(key)) or b"\x80\x05}\x94.")
-    task_status = LoadStatus(data.get("status", 5))
+    raw = await Cache.get(key)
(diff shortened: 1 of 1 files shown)


antarcticrainforest (changes requested):
This doesn't make much sense. Your simply saying that any entry that doesn't exist is waiting. 

So if I request value `foo.bar` the server will tell me it's waiting for `foo.bar` to become available. 

Also the value is not `gone` it's more like `unkown`. Because it tells you, you were requesting something that's not cached (either because it has never been, or is gone). So it can't tell much about it. Hence "unkown". 

If you really want to achieve a solution to your problem you'd have to add a semaphore or something to the k/v store after or before the load instruction was sent to the broker in `freva_rest.freva_data_portal.utils.publish_dataset`. Something that tells the system this job has been queued but hasn't been picked up by a worker yet.

PR #332, opened 2026-04-10
feat: Add flavour concept into STACAPI and enrich the collection via an static core in Solr

tropicrainforest:
This PR introduces two main changes.

As first feature, it brings the concept of translators (“flavours”) into the STAC API. The main goal is to support a request from the ESGF group. They want to deploy Freva, and as a result the STAC API, on their side while continuing to use the Solr search engine but with a different `managed_schema.xml`. With this PR, after deployment they can define a new global flavour called `ESGF` and map the existing `freva` facets to ESGF facets without any need to have new managed_schema or modify anything there. This would also let them to expose a STAC API endpoint such as `/api/freva-nextgen/stacapi/esgf`.

But still there is one important caveat that we couldn't find a solid solution for: the number of facets currently available in our main `managed_schema.xml` is smaller than the number of metadata fields used by `ESGF`. ESGF has roughly 58 metadata keys, which is more than what we currently expose in our managed schema. So even with translation in place, we still cannot represent all ESGF metadata one-to-one. On the positive side, a quick comparison shows that some ESGF facets may not be strictly necessary for our use case, and some are simply represented differently. For example, ESGF defines the bounding box as four separate facets, while in Freva we represent it as a single `bbox` field. There are several similar cases like this. Hopefully after a discussion with them we probably can get to a conclusion on the number of keys.

As a second change, for the Waterfall project, this PR adds support for reading constant collection metadata from a `static` Solr core, as introduced in [freva-service-config PR #85](https://github.com/freva-org/freva-service-config/pull/85). This makes catalog curation much easier and let the process to be handled automatically. If that PR is not approved or the `static` core cannot be reached, this PR still works correctly because it automatically falls back to the default values that were already defined before.

Also the other benefit is, once [freva-service-config PR #85](https://github.com/freva-org/freva-service-config/pull/85) is approved, we can also add a new table to the nightly automatic crawler (`-general`) for our currently running instances, so that updated `static` metadata can be crawled and ingested automatically into the deployed freva instances and STACAPI can have thumbnails and long description and so on.

### DIFF
--- a/dev-env/static-managed-schema.xml
+++ b/dev-env/static-managed-schema.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<schema name="static" version="1.6">
+  <uniqueKey>id</uniqueKey>
+
+  <!-- Standard Solr field types required by the default configset -->
+  <fieldType name="binary"      class="solr.BinaryField"/>
+  <fieldType name="boolean"     class="solr.BoolField"        sortMissingLast="true"/>
+  <fieldType name="booleans"    class="solr.BoolField"        sortMissingLast="true" multiValued="true"/>
+  <fieldType name="pint"        class="solr.IntPointField"    docValues="true"/>
+  <fieldType name="pints"       class="solr.IntPointField"    docValues="true" multiValued="true"/>
+  <fieldType name="plong"       class="solr.LongPointField"   docValues="true"/>
+  <fieldType name="plongs"      class="solr.LongPointField"   docValues="true" multiValued="true"/>
+  <fieldType name="pfloat"      class="solr.FloatPointField"  docValues="true"/>
+  <fieldType name="pfloats"     class="solr.FloatPointField"  docValues="true" multiValued="true"/>
+  <fieldType name="pdouble"     class="solr.DoublePointField" docValues="true"/>
+  <fieldType name="pdoubles"    class="solr.DoublePointField" docValues="true" multiValued="true"/>
+  <fieldType name="pdate"       class="solr.DatePointField"   docValues="true"/>
+  <fieldType name="pdates"      class="solr.DatePointField"   docValues="true" multiValued="true"/>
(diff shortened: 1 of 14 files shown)


freva-rest/src/freva_rest/stac_api/core.py
```diff
@@ -54,14 +57,24 @@ def __init__(
         self,
         config: ServerConfig,
         *,
+        flavour: str = "freva",
         limit: int = 12,
         token: Optional[str] = None,
         datetime: Optional[str] = None,
         bbox: Optional[str] = None,
         uniuq_key: Literal["file", "uri"] = "file",
+        _translator: Any = None,
```
antarcticrainforest, line 66: wrong type.

freva-rest/src/freva_rest/stac_api/core.py
```diff
@@ -274,116 +291,196 @@ async def get_landing_page(self) -> Dict[str, Any]:
...
+        title = (static or {}).get("title") or collection_id.upper()
+        description = (
+            (static or {}).get("description")
+            or f"Collection {collection_id.upper()}"
+        )
+        license_id = (static or {}).get("license") or "proprietary"
+        license_url = (
+            (static or {}).get("license_url")
+            or "https://opensource.org/license/bsd-3-clause"
+        )
```
antarcticrainforest, line 354: Something doesn't add up here. bsd-3 is not proprietary? If there isn't any license attached then don't add any!

dev-env/static-managed-schema.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<schema name="static" version="1.6">
```
antarcticrainforest, line 2: Do we need this file if it comes from the service config?

antarcticrainforest (changes requested):
I am mainly concerned about the naming of the core. I find the name "static" not really informative. Will this exclusively used for stac catalogues?

antarcticrainforest:
I am going to close this for now!

PR #334, opened 2026-04-17
fix: pass app_name as keyword argument to TokenStore

tropicrainforest:
`TokenStore(self.app_name)` was passing "freva" as the `path` argument instead of `app_name`, causing tokens to be written to a file literally named `freva` in the current working directory rather than to
`~/.cache/token-store.json`.

This meant the token cache was never found on subsequent calls, forcing a new browser-based auth flow every time and breaking non-interactive use cases.


antarcticrainforest (approved):
Good catch, I was just about to look for the cause of the weirdness! Thanks!!!!

LGTM!

PR #336, opened 2026-04-17
Replace IDP token passthrough with RS256 token broker (RFC 8693)

antarcticrainforest:
### Summary

This PR replaces direct IDP token passthrough with a token broker architecture. The REST API now issues its own RS256-signed freva JWTs to clients and handles all IDP communication internally. Clients never see or store IDP tokens. This is in line with current OAuth specifications. The  passthrough approach violates RFC 7519 §4.1.3 (tokens must only be accepted by their intended audience) and the least privilege guidance in RFC 9700, since clients are receiving and storing IDP tokens scoped to the IDP rather than tokens scoped to the freva API. The new broker pattern aligns with RFC 8693's token exchange model, where the REST API acts as a trusted intermediary that exchanges IDP credentials for service-scoped tokens.

### Architecture

The auth flow now works as follows:

1. Client completes any standard auth flow (device code, auth code) via the py-oidc-auth endpoints
2. Client POSTs the resulting code or device-code to `POST /auth/v2/token`
3. The API calls the IDP internally, validates the IDP token, mints a freva RS256 JWT (`aud=freva-api`), stores the IDP refresh token in MongoDB, and returns the freva JWT
4. Clients use the freva JWT on all protected endpoints
5. When the freva JWT nears expiry the client POSTs it back as `refresh-token` — the API silently exchanges the stored IDP refresh token for fresh IDP tokens, rotates the session, and returns a new freva JWT

### Future changes

Since the minted tokens are note standardized we actually don't need a `userinfo` endpoint anymore. Tokens can be validated and their payload inspected entirely on client side. Checking via `userinfo` only adds unnecessary traffic and latency, since the server is doing exactly the same as the client  - hence a direct validation would save the round trip via the api.  @gekinci feel free to give your opinion on the topic.


freva-rest/src/freva_rest/auth/__init__.py
```diff
@@ -0,0 +1,356 @@
...
+
+# ---------------------------------------------------------------------------
+# Unified token endpoint
+# ---------------------------------------------------------------------------
+
+
+@auth_router.post(
+    "/auth/v2/token",
+    response_model=Token,
+    tags=["Authentication"],
```
tropicrainforest: please remove tag since router inherits `Authentication` from py-oidc-auth already by default

freva-rest/src/freva_rest/auth/__init__.py
```diff
@@ -0,0 +1,356 @@
...
+
+# ---------------------------------------------------------------------------
+# Supporting endpoints
+# ---------------------------------------------------------------------------
+
+
+@auth_router.get(
+    "/auth/v2/.well-known/jwks.json",
+    response_class=JSONResponse,
+    tags=["Authentication"],
```
tropicrainforest: here as well

freva-rest/src/freva_rest/auth/__init__.py
```diff
@@ -0,0 +1,356 @@
...
+    sub: str
+    exp: int
+    email: Optional[str] = None
+
+
+# Freva-specific endpoints (not provided by py-oidc-auth)
+@auth_router.get(
+    "/auth/v2/.well-known/openid-configuration",
+    response_class=JSONResponse,
+    tags=["Authentication"],
```
tropicrainforest: here as well

tropicrainforest (changes requested):

antarcticrainforest:
I have a question about the audience. Currently we fix the audience. Meaning that any freva token is valid across instances. We could of course restrict things using claim checks. Or we make the audience configurable. 

I guess making audience configurable with a fallback could be it?

tropicrainforest:
> I have a question about the audience. Currently we fix the audience. Meaning that any freva token is valid across instances. We could of course restrict things using claim checks. Or we make the audience configurable.
> 
> I guess making audience configurable with a fallback could be it?

Making the audience configurable makes sense in principle, but there are two problems with it in practice.

1. First, a fallback to `None` has no value. If `oidc_audience` is not set, audience verification is simply skipped; which is fine as a conscious opt-out, but it should be treated as such: operators who don't set it simply don't want this level of security, and that's a valid choice. But when one chooses wrong, then one should get ValueError to be aware and change it.

2. Second, the configurable approach is fragile. Based on how Keycloak issues tokens, the client name appears in `azp`; not in `aud`. In the token below, the client is `codes`, but `"codes"` does not appear anywhere in `aud`:

```json
{
  "aud": ["xces", "realm-management", "regiklim", "freva", "sp9-test-freva", "account"],
  "azp": "codes"
}
```

So an operator who naively sets `oidc_audience = "codes"` will break all token validation, because PyJWT will look for `"codes"` in `aud`, fail to find it, and reject every token. Making it configurable requires operators to understand the internal structure of their Keycloak tokens and know to look at `aud` rather than `azp`; which is easy to get wrong silently.

The configurable approach only works reliably with clear documentation that the value must match an entry actually present in `aud` (e.g. `"sp9-test-freva"`, not `"codes"` in above example). Given this complexity, sticking to claim-based restrictions is more reliable and straightforward for this use case, since claims like `realm_access.roles` or `resource_access` are more predictable and easier to reason about than the `aud` field.

antarcticrainforest:
No, that's not what I meant. We don't do any aud checks against the IDP. The IDP token is solely used for authentication not for authorisatio. We use our own tokens for authorisation. Hence I am talking about authorisation of the endpoints.

Fallback wouldn't be None, it would be what I have set now, freva-rest or whatever it was.

If two rest-api's like nextgems and the one at JSC (which they have finally managed to set up) share the same audience (such as warmworld), users logged in via our nextgems could use the zarr streaming endpoint at JSC and vice versa. This is the clue about token minting. We are giving out tokens and decide when they are valid and to whom. So it means we decide about federation! That's the clue about this!

antarcticrainforest:
I was just thinking about federation a little. We would also need to configure the trusted issuers `iss`. And then also fetch the kids. This should be done periodically in case the tokens get rotated on the other side. 

I am going to implement that. Because I think freva federation is a cool feature.

tropicrainforest:
I thought a bit more about this, and I think it is a bit tricky. I feel like using only `aud` does not seem enough to make this properly federated:

1. For example, DKRZ gives user X a token with `aud="warmworld"`. X then sends it to JSC Zarr. JSC accepts it just because the `aud` matches, even though the token was not specifically issued for JSC. We could verify a few more claims to make this safer, but even then it still feels more like a workaround than a well-defined federation model, especially if it is not based on a clear written specification. And I'm not sure if there a written specification for this purpose.

2. DKRZ may allow user X because X is a WarmWorld member and registered there and accepted the data policy of DKRZ while regestring. But JSC may have stricter rules for access to its own data with its own data policy, even if the data belongs to the same project. If JSC still accepts X’s token only because it sees `aud="warmworld"` and knows X is part of WarmWorld, then JSC’s local policy is effectively bypassed. This is the risky part: DKRZ would be making the authorization decision on behalf of JSC. Look into this model:

user X -> py-oidc-auth router -> DKRZ keyclock -> Freva token issuer -> JSC token receiver -> access to JSC

As you see JSC OIDC policy is not playing any role in giving access to user X to the data that are located in its center.

3. If a token leaks somewhere, it could potentially be used against more than one data center if all of them accept the same broad audience. That may be a somewhat pessimistic scenario, but it is still part of the risk.

I believe as long as we don't follow any written specification for this, there is not safe approach for this, but a safer approach would be to make the token more specific by describing the exact action of user (project name, only zarr steamin, ro operation, and things like this) in different arguments and also target and specify destination, and then let the receiving site apply its own policy check(I don't know how tbh). Since it's about security, I think we should play more skeptic here.

antarcticrainforest:
It's not *solely* based on `aud` - that would be wrong. We can't also check `claims`, claims are the wrong place for this. 

We have an issuer with the token. And the issuer must match a collection of *pre*-defined allowed issuers . This is where an admin comes into play. They must explicitly allow that. The token itself is validated with the `kid` of the token. This procedure of token federation is indeed inline with *OpenID Connect Federation* - as far as I can tell not an RFC yet - only draft.

Where it gets a little murky is the token broker itsefl (RFC 8693): 
RFC 8693 defines a standard `grant_type=urn:ietf:params:oauth:grant-type:token-exchange` which we are not using. We're using a custom flow that is different. In RFC 8693 the client presents a `subject_token` and receives a new one.

But I'd rather not have a discussion about internet engineering task force specs and their requests for comments. 

If you really want to go that route then we *mustn't* implement any authentication on our side. Clients will always have to directly authenticate with the IDP not us. I don't think that this is what we want.

freva-rest/src/freva_rest/auth/__init__.py
antarcticrainforest: Because then it appears twice in the help? right...removed!

freva-rest/src/freva_rest/auth/__init__.py
antarcticrainforest: removed!

freva-rest/src/freva_rest/auth/__init__.py
antarcticrainforest: removed

antarcticrainforest:
I was just thinking, that maybe this whole machinery would fit better in `py-oidc-auth` ? Thoughts @tropicrainforest ?

tropicrainforest:
> I was just thinking, that maybe this whole machinery would fit better in `py-oidc-auth` ? Thoughts @tropicrainforest ?

My understanding is that `py-oidc-auth` is a more generic library focused on OIDC authentication itself. It helps with things like handling authentication flows, validating claims, and exposing OIDC-related functionality in a REST-friendly way on top of standard OIDC capabilities.

On the otjher side, a token broker, adds another layer on top. It receives or validates IdP-issued tokens, mints service-specific tokens, stores refresh state, and based on latest commit, it introduces federation and trust relationships between deployments. That is no longer just "OIDC authentication support"; it becomes application-level token management and cross-service trust handling. and also it doesn't follow the naming of the package anymore.
i would say let's keep `py-oidc-auth` mainly focused on authentication against an OIDC provider, and keep token brokering in a separate repo. That would make the separation of responsibilities much clearer: `py-oidc-auth` would remain about authentication, while the token broker would handle local token issuance, delegation, session storage, and possibly federation between services.
Because, 
1. it keeps ` py-oidc-auth` simpler and more focused. Not every operator who wants OIDC authentication also wants token brokering or federation. If we put everything into one package, users may end up with much more functionality than they actually need. Based on our experience with Freva-legacy, a more decentralized design with smaller, focused components is often easier to understand, easier to adopt, and more attractive than delivering everything as one large block.
2. a token broker is not just a lightweight library feature. It usually requires running a service, managing signing keys, and storing session or refresh-token state. That makes it operationally heavier than a lean pip-installable OIDC helper. If this complexity is built directly into `py-oidc-auth`, some operators may find it too heavy for their use case and may prefer another solution or software instead.

freva-rest/src/freva_rest/auth/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+    idp_token_obj: Token, expiry_seconds: int = TOKEN_EXPIRY_SECONDS
+) -> Token:
+    """Validate an IDP Token, mint a freva JWT, persist the session."""
+    try:
+        idp_claims: IDToken = await auth._get_token(
+            idp_token_obj.access_token,
+            effective_claims=server_config.oidc_token_claims or None,
+        )
+    except InvalidRequest as exc:
+        raise HTTPException(status_code=exc.status_code, detail=exc.detail)
```
tropicrainforest, line 318: A thought for handling guests and third-party IDPs like Google and GitHub which are not part of any claim: instead of blocking at token issuance, we catch `InvalidRequest` and fall back to `roles=[]`? what do you think?

Users outside the configured claim, or those authenticating via third-party IDPs, still receive a valid token but with no roles attached. The 403 is then raised in `token_wrapper` when they hit a protected endpoint(for example zarr loading).
Via this approach, anyone can get a token, but access is enforced at the endpoint level. I think it can keep a clean separation between authentication and authorization

```suggestion
    roles = idp_claims.flattened_roles
except InvalidRequest:
    idp_claims = IDToken.from_token(idp_token_obj.access_token)
    roles = []
```

and in `token_wrapper`:

```python
def token_wrapper(*required_roles: str) -> Callable[..., Awaitable[IDToken]]:
    async def _dep(
        authorization: Annotated[Optional[str], Header()] = None,
    ) -> IDToken:
        claims = await check_token(authorization, required_roles)
        if not claims.roles:
            raise HTTPException(status_code=403, detail="Guest access not permitted.")
        return claims
    return _dep
```

tropicrainforest (commented):
Since we have dropped the `/systemuser` endpoint, how about, user claims from OIDC decides about the permission of the guest users. It's a new discussion around guest users

antarcticrainforest:
This was introduced because:

* G refused to add `aud=foo` to the IDP tokens because those tokens are issued for Keycloak itself, not for freva specifically,  adding a freva-specific audience claim to an IDP token is semantically wrong.
* But without an aud claim, passing IDP tokens through to freva-rest and doing audience validation there is meaningless, we'd either have to skip aud validation entirely (security hole) or accept any token from that G's Keycloak realm regardless of what it's for
* RFC 7519 requires that if aud is present, the recipient must verify it,  and if it's absent,we lose a key defense against token misuse across services.


G was right and I think the broker is the right architectural respons. The IDP issues tokens for itself, the broker exchanges them for freva-scoped tokens with `aud=foo`, and freva-rest only accepts tokens it issued itself.

I think the broker pattern is generic enough that other projects hitting the same G problem would benefit from it.
About additional services: We could define multiple token stores, such as mongo, postgres, mysql and sqlite. All except mongo would be easy to cover with sqalchemy adding a mongo-client isn't too bad.

antarcticrainforest:
Closing this now. I'll submit a PR for `py-oidc-auth` shortly

PR #337, opened 2026-04-21
Token federation

antarcticrainforest:
This is a rather small PR that adds the latest updates of `py-oidc-auth` to include a broker mode and establish and optional network for trust for token federation. 

Most of the changes were done in the tests.


freva-rest/src/freva_rest/auth.py
```diff
@@ -19,6 +19,12 @@
     scopes=server_config.oidc_scopes,
     proxy=server_config.proxy,
     claims=server_config.oidc_token_claims or None,
+    broker_mode=True,
```
tropicrainforest, line 22: nitpick: do you think, having a boolean `broker_mode` arg is needed here? How about, as soon as we are using `broker_*` arg, it can recognize we are demanding broker along with OIDC?

tropicrainforest (approved):
LGTM!

freva-rest/src/freva_rest/auth.py
antarcticrainforest: That's not an issue with this service

antarcticrainforest: If, then it should go into `py-oidc-auth`. I am not a fan of implicit switches.

PR #349, opened 2026-04-28
User permissions check

antarcticrainforest:
## Filesystem permission checks for data portal access

This PR adds read-permission enforcement for zarr publishing and pre-signed URL creation. Before this change, any authenticated user could trigger serialization of data they wouldn't normally be able to read on the filesystem.

### How it works

- The REST API extracts the authoritative local username from the IDP's userinfo response (configurable via `systemuser_claim`, supports JMESPath for nested claims like `foo.bar.username`)
- On publish or presign, the API sends an `access_check` message to the data-loader via the Redis broker
- The data-loader performs a filesystem permission check (`stat`-based when non-root, `su + test -r` when root) and replies
- Results are cached in Redis for 5 minutes keyed by `username:path_hash`
- Users without a local username (federated IDP logins, missing claim) are treated as guests and may only access world-readable data
- Metadata endpoints for private zarr stores also check permissions on first access

### Other changes

- `@background` decorator on `from_object_path` and `get_zarr_chunk` in the data-loader. Without this, heavy serialization tasks block the message loop and cause access-check timeouts. This is a pragmatic workaround. Long-term we should consider an async event loop or separate control channel.


tropicrainforest (approved):
LGTM!
I took a test and it seems everything that is needed for web client, has been properly implemented here. After merging this we would open the web PR to get whole the isguest + data-loader off the desk.

tropicrainforest:
@antarcticrainforest 
One case I overlooked in `xarray-prism` is when the requested data doesn’t exist. I took a note to fix this before releasing the new version of data-loader.
At the moment, there’s no clear indication of whether the data actually exists. Instead, during the detection stage, it throws an error saying it couldn’t detect the format, which isn’t very informative or helpful.
For example:
```
Xarray Prism: cannot detect format for '/arch/bb1203/freva_test/observations/grid/CPC/CPC/cmorph/30min/atmos/30min/r1i1p1/v20210618/pr/pr_30min_CPC_cmorph_r1i1p1_201609020000-201609020030.nc'
  Specify manually:   xarray.open_dataset(uri, engine='ENGINE_NAME')
```
It should directly says, `the data couldn't be found`
we could merge the current PR now, but it would be better to delay the release of data-loader until the fixed version of `xarray-prism` is available.

antarcticrainforest:
There is one more thing, that might be worth while fixing for prism. cfgrib creates an index file next to the data so that next time the data can be opened faster. We "don't" want that. In principle we are root so we can write anywhere.  So the index would just end up next to the data - as root user. Which would be a really bad thing.

I think for backend_kwargs for cfgrib you could use an env variable. I don't think you want to disable it but we could set an env variable and data-loader isn't writing it or writing to the cache.


Also the log information in prism are kind of annoying can you maybe disable or at least make them debug.

```

```There is one more thing, that might be worth while fixing for prism. cfgrib creates an index file next to the data so that next time the data can be opened faster. We "don't" want that. In principle we are root so we can write anywhere.  So the index would just end up next to the data - as root user. Which would be a really bad thing.

I think for backend_kwargs for cfgrib you could use an env variable. I don't think you want to disable it but we could set an env variable and data-loader isn't writing it or writing to the cache.

PR #351, opened 2026-05-11
hotfix: enforce system user access gate in systemuser endpoint

tropicrainforest:
While adjusting the web guest user flow, a potential security hole for third-party IDP users came to light. The `systemuser` endpoint was returning 200 for any authenticated user regardless of whether they matched the configured `oidc_systemuser_claim`. A 403 guard has been added when the JMESPath expression returns no match, so only users satisfying the claim condition (e.g. membership in the `ch1187` group) are recognized as system users. If `oidc_systemuser_claim` is not set, everyone is treated as a regular user.
This doesn't affect filesystem-level access on the crucial data-loader, since usernames are verified via `pwd` there and `systemyuser` has not been used there, but it adds a guard at the web layer, e.g. for the plugin submission button.

### DIFF
--- a/freva-client/src/freva_client/query.py
+++ b/freva-client/src/freva_client/query.py
@@ -1275,7 +1275,7 @@ class databrowser:
             overview["Note"] = note
         overview["Available search flavours"] = overview.pop("flavours")
         overview["Search attributes by flavour"] = overview.pop("attributes")
-        return yaml.safe_dump(overview, sort_keys=False)
+        return cast(str, yaml.safe_dump(overview, sort_keys=False))
 
     @property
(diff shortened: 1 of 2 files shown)


antarcticrainforest:
Do we really want this? I am very doubtful about this whole endpoint at all.

tropicrainforest:
@antarcticrainforest since we don't use `systemuser` endpoint on the web-client anymore, shall we ditch the endpoint in in this PR or we need to open another PR for?

antarcticrainforest:
Let's keep it, maybe we will need it in the future

antarcticrainforest (approved):
LGTM

PR #361, opened 2026-05-18
perf: add adaptive facet.method for faster Solr faceting

tropicrainforest:
Benchmarking shows that Solr's default `facet.method`, `fc` carries a huge cost on unfiltered queries when the field cache is cold; the first request after a restart pays the full price of building the uninverted index.
Switching to `enum` for unfiltered queries avoids this entirely. Again based on the benchmark, for filtered queries, `fc` remains the better choice as it scales with the result set and becomes very fast once filters narrow the match.

A quick benchmark on `gems.dkrz.de` after a cold cache:

```bash
curl "http://gems.dkrz.de:8983/solr/admin/cores?action=RELOAD&core=latest"
SOLR="http://gems.dkrz.de:8983/solr/latest/select"
for method in enum fc; do
  for fq in "" "project:(cmip6)" "project:(cmip6)&fq=variable:(tas)"; do
    curl -w "method=$method fq='$fq': %{time_total}s\n" -o /dev/null -s \
      "$SOLR?q=*:*&rows=0&facet=true&facet.limit=-1&facet.mincount=1&facet.method=$method&facet.field=variable&facet.field=model&facet.field=ensemble&facet.field=project&facet.field=user&fq=$fq"
  done
  echo "---"
done
```

```
method=enum fq='': 0.663809s
method=enum fq='project:(cmip6)': 0.683745s
method=enum fq='project:(cmip6)&fq=variable:(tas)': 0.659431s
---
method=fc fq='': 2.203488s
method=fc fq='project:(cmip6)': 0.341317s
method=fc fq='project:(cmip6)&fq=variable:(tas)': 0.027627s
```

`enum` is 3 times faster unfiltered; on the opther hand `fc` is 20 times faster on a narrow filter. This change is already live at https://freva-deployment.dkrz.de/databrowser/


antarcticrainforest (approved):
LGTM!

PR #362, opened 2026-05-19
perf(api): cache extended Solr search results

antarcticrainforest:
Add an async TTL cache for extended-search responses to avoid repeated expensive Solr facet queries during databrowser page loads.

The cache is keyed by the search parameters and skipped for zarr-stream requests to avoid caching side-effecting operations. Cached values are deep-copied on read/write to prevent accidental mutation of shared cache state.

The cache can also be used to query system user names.


freva-rest/src/freva_rest/rest.py
```diff
@@ -87,19 +88,61 @@
...
+    stop_event: asyncio.Event,
+    interval_seconds: int,
+) -> None:
+    """Refresh the extended-search cache until shutdown is requested."""
+    from .databrowser_api.core import Solr
+
+    while not stop_event.is_set():
+        try:
+            for key in ("file", "uri"):
+                await Solr.refresh_extended_search_cache(uniq_key=key, max_results=100)
```
tropicrainforest, line 101: I think both sides should agree on the same value. Either raise the warmer here to `max_results=150` to match the extended-search endpoint default, or lower the endpoint default to 100 to match the warmer. otherwise they never share the same cache key and the warmer is useless. Yes, of course user can change max-result to other number by themselves, and as a result they get new cache key, but this is more about the better experience for users on the data-browser landing page

freva-rest/src/freva_rest/databrowser_api/core.py
```diff
@@ -784,9 +855,20 @@ async def extended_search(
...
+            uniq_key=self.uniq_key,
+            query_items=query_items,
+        )
+        cached: Optional[Tuple[int, Dict[str, Any]]] = await self._cache.get(key)
+        if use_cache is False or cached is None:
+            async with self._session_get() as res:
+                search_status, search = res
+        else:
+            search_status, search = cached
+        if set_cache:
```
tropicrainforest: Am I wrong or `set_cache` always writes, even on a cache hit? That would reset the TTL every time and keep stale data alive forever. Not 100 percent sure

freva-rest/src/freva_rest/config.py
```diff
@@ -36,7 +40,42 @@
 
 ConfigItem = Union[str, int, float, None]
 BUILTIN_FLAVOURS = ["freva", "cmip6", "cmip5", "cordex", "user"]
+SEARCH_CACHE: TTLCache[str, Any] = TTLCache(
+    maxsize=1024,
+    ttl=600,
+)
+EXTENDED_SEARCH_LOCK = asyncio.Lock()
+KEY_LOCKS: dict[str, asyncio.Lock] = {}
 T = TypeVar("T", str, int)
```
tropicrainforest, line 47: dead lines?

freva-rest/src/freva_rest/databrowser_api/core.py
```diff
@@ -784,9 +855,20 @@ async def extended_search(
...
+        query_items = normalise_solr_query(self.query)
+        key = hash_from_query(
+            url=self.url,
+            uniq_key=self.uniq_key,
+            query_items=query_items,
+        )
+        cached: Optional[Tuple[int, Dict[str, Any]]] = await self._cache.get(key)
+        if use_cache is False or cached is None:
+            async with self._session_get() as res:
+                search_status, search = res
```
tropicrainforest, line 875: I'm not sure what behavior we will experience here on a scenario that we receive `res.status != 200`

freva-rest/src/freva_rest/config.py
```diff
@@ -36,7 +40,42 @@
...
+G = TypeVar("G")
+
+
+class AsyncTTLCache(Generic[G]):
+    """Small async-safe wrapper around a TTL cache."""
+
+    def __init__(self) -> None:
+        self._cache = SEARCH_CACHE
+        self._lock = asyncio.Lock()
```
tropicrainforest: nitpick: generally asyncio is single-threaded and doesn't harm in practice. I'm askign just to learn: couldn't we share this lock globally instead of creating the lock for each instance? it means they all share the same underlying SEARCH_CACHE

freva-rest/src/freva_rest/auth.py
```diff
@@ -12,11 +13,12 @@
 from py_oidc_auth.utils import get_userinfo
 from pydantic import BaseModel, Field
 
-from .config import ServerConfig
+from .config import AsyncTTLCache, ServerConfig
 
 logger = logging.getLogger(__name__)
 
 server_config = ServerConfig()
+cache: AsyncTTLCache[str] = AsyncTTLCache()
```
tropicrainforest: hmm, when we import AsyncTTLCache from config and calling it here, are we using the same SEARCH_CACHE for both auth and extended-search? If so, a heavy auth load could push warm search entries out and the other way around. Maybe I'm being too pessimistic here, but it feels like they should have their own separate caches just to be safe.

tropicrainforest (commented):
My favourite topic, caching; great work Martin, this already makes the endpoint incredibly faster. Thanks really
One main comment for maintainability and future reuse: could we move the caching logic into a small shared utility/helper instead of keeping it directly inside `extended_search`?
Right now the cache key generation, cache lookup, Solr fallback, cache write, and refresh behaviour are tightly coupled to this method. If we later want to cache other read-heavy endpoints, we may end up duplicating the same logic or fixing cache-related bugs in multiple places.
Maybe we could introduce something like a reusable `get_or_fetch_cached` helper in `utils` or a small cache service, where each caller the need arguments like `key_parts`, `use_cache`, `ttl`, `should_cache condition`, `fetch callback` and so on.
Then `extended_search` can stay focused on building the Solr query and post-processing the response, while the caching behaviour remains reusable for future endpoints.

freva-rest/src/freva_rest/rest.py
antarcticrainforest: Not sure what you mean here, but the web has 100.

freva-rest/src/freva_rest/databrowser_api/core.py
antarcticrainforest: true

freva-rest/src/freva_rest/databrowser_api/core.py
antarcticrainforest: right

freva-rest/src/freva_rest/config.py
antarcticrainforest: done

freva-rest/src/freva_rest/auth.py
antarcticrainforest: In principle it doesn't hurt but I'll implement something.

antarcticrainforest:
> My favourite topic, caching; great work Martin, this already makes the endpoint incredibly faster. Thanks really One main comment for maintainability and future reuse: could we move the caching logic into a small shared utility/helper instead of keeping it directly inside `extended_search`? Right now the cache key generation, cache lookup, Solr fallback, cache write, and refresh behaviour are tightly coupled to this method. If we later want to cache other read-heavy endpoints, we may end up duplicating the same logic or fixing cache-related bugs in multiple places. Maybe we could introduce something like a reusable `get_or_fetch_cached` helper in `utils` or a small cache service, where each caller the need arguments like `key_parts`, `use_cache`, `ttl`, `should_cache condition`, `fetch callback` and so on. Then `extended_search` can stay focused on building the Solr query and post-processing the response, while the caching behaviour remains reusable for future endpoints.

I kind of disagree here. The cache lookup is general it lives in settings we use it in so that's fine. The hash key generation for is custom for solr queries this is why it's there. I decided to make it a lru_cache because then it's faster. But that has the caveat that this needs to stay solr specific. Otherwise I'd have to create a recursive function that creates immutable options from `**kwargs` to be able to use lru_cache. In this case the speed gain of lru cache would be gone.

freva-rest/src/freva_rest/rest.py
tropicrainforest: yes, true
the warmer works today because freva-web happens to use 100 https://github.com/freva-org/freva-web/blob/main/assets/js/Containers/Databrowser/constants.js#L28, but that creates a hidden dependency. If the frontend ever changes that number, the warmer silently stops helping without any obvious indication. I was thinking if we kept `max_results` here in sync with the extended-search default one, at least any smoke run would still benefit any new web client that is going to be built via this public endpoint. But I was just afraid it becomes a magic number. Anyway, we can keep this as is for now.

tropicrainforest:
> > My favourite topic, caching; great work Martin, this already makes the endpoint incredibly faster. Thanks really One main comment for maintainability and future reuse: could we move the caching logic into a small shared utility/helper instead of keeping it directly inside `extended_search`? Right now the cache key generation, cache lookup, Solr fallback, cache write, and refresh behaviour are tightly coupled to this method. If we later want to cache other read-heavy endpoints, we may end up duplicating the same logic or fixing cache-related bugs in multiple places. Maybe we could introduce something like a reusable `get_or_fetch_cached` helper in `utils` or a small cache service, where each caller the need arguments like `key_parts`, `use_cache`, `ttl`, `should_cache condition`, `fetch callback` and so on. Then `extended_search` can stay focused on building the Solr query and post-processing the response, while the caching behaviour remains reusable for future endpoints.
> 
> I kind of disagree here. The cache lookup is general it lives in settings we use it in so that's fine. The hash key generation for is custom for solr queries this is why it's there. I decided to make it a lru_cache because then it's faster. But that has the caveat that this needs to stay solr specific. Otherwise I'd have to create a recursive function that creates immutable options from `**kwargs` to be able to use lru_cache. In this case the speed gain of lru cache would be gone.

Agreed

tropicrainforest (approved):
LGTM!

PR #363, opened 2026-05-25
Hardening

antarcticrainforest:
## Summary

This PR strengthens the data-portal loading path and improves operational behaviour for running the data-loader worker on HPC infrastructure.

The main change is that permission checks are now enforced not only by the REST API before publishing a load request, but also by the data-loader worker when it receives a `uri` broker message. This means that even if a message is published to the worker directly, bypassing the REST API, the worker still verifies that the resolved system user is allowed to read the requested input paths before loading data.

The data-loader no longer starts a distributed Dask cluster. This reduces operational complexity, avoids exposing a Dask scheduler/dashboard from the worker process, and makes the service easier to reason about from a security perspective.

## Changes

- Add worker-side permission checks for `uri` broker messages before loading data.
- Pass the resolved system username through the broker message into the worker.
- Add a dedicated permission-denied load state for failed access checks.
- Remove the distributed Dask `LocalCluster` usage from the data-loader worker.
- Add bounded Redis connection pooling using `BlockingConnectionPool`.
- Redact sensitive Redis connection fields in logs.
- Improve Redis/cache error handling and invalid-token behaviour.
- Locally cache loaded data xarray data sets
- Enable mulit processing for the data-loader
- Sanitize broker messages before processsing
- Improve logging behaviour in tests by falling back to plain stream logging where appropriate.
- Extend tests for access control, Redis cache reads, retry/reload behaviour, and logger handling.

## Security impact

This adds a second authorization gate at the worker boundary. The REST API still performs the normal authentication and permission checks, but the worker now also validates read access before opening source data. This provides defense in depth in case a broker message is submitted without going through the REST API.

The data-loader remains responsible only for loading data that the authenticated system user is allowed to read. Chunk requests operate on already-authorized cached datasets and do not perform expensive filesystem checks for every chunk.

## Operational impact

Removing distributed Dask simplifies the runtime model and reduces exposed network surface. Redis connections are now bounded by an explicit connection pool, which prevents unbounded connection growth under load.

Sensitive Redis configuration values are no longer printed in logs.


freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -431,60 +462,78 @@ def load_object(self, key: str) -> Tuple[Dict[str, Any], Dict[str, xr.Dataset]]:
...
-        dset_cache = self.cache.get(f"{key}-dset")
-        if metadata_cache is None or dset_cache is None:
-            raise KeyError(f"{key} uuid does not exist (anymore).")
-        load_dict: LoadDict = cloudpickle.loads(metadata_cache)
-        dsets = cast(Dict[str, xr.Dataset], cloudpickle.loads(dset_cache))
-        data_logger.debug("Loading %s ... done", key)
-        return cast(Dict[str, Any], load_dict["data"]), dsets
+        result = self._cache_lookup(key)
+        if result is not None:
+            return result
```
tropicrainforest, line 467: Not sure I've tracked this properly but it seems the caching/eviction across workers leaks stale data to new chunk requests.

Local cache hits return at `self._cache_lookup(key)` without ever calling Redis (the `self.cache.get` calls are below the early return, so they only run on a miss, which makes sense when there's only one worker). That means a worker can keep serving a dataset after Redis has expired it or another worker has reloaded it, because `self.cache.brpop(channel, timeout=1)` delivers each message to a single consumer, so `_evict_object_cache` runs in that one process and nowhere else.

For example, on a machine with two workers, w1 and w2:

1. Chunk request for TOKEN -> `brpop` delivers to w1 -> `load_object(TOKEN)` -> `_cache_lookup` miss -> `self.cache.get` populates `w1._object_cache[TOKEN]`. At the end: Redis = {TOKEN, TOKEN-dset}, w1 = {TOKEN}, w2 = {}.
2. Another chunk request for TOKEN -> `brpop` delivers to w2 -> same miss path -> populates `w2._object_cache[TOKEN]`. Now: Redis = {TOKEN, TOKEN-dset}, w1 = {TOKEN}, w2 = {TOKEN}.
3. Now a reload via `Cache.lpush("data-portal", {"uri": {..., "reload": True}})` -> `brpop` delivers to w1 only (destructive pop) -> `redis_callback` -> `spawn` -> `from_object_path` -> `_evict_object_cache(TOKEN)` clears w1 only -> re-reads source, writes new TOKEN and TOKEN-dset to Redis. At the end: Redis = {TOKEN', TOKEN-dset'}, w1 = {}, w2 = {TOKEN <- stale}.
4. Chunk request -> `brpop` delivers to w2 -> `_cache_lookup` HITS, returns without touching or asking Redis -> `get_zarr_chunk` encodes from the old dataset -> `self.cache.setex(f"{TOKEN}-{var}-{chunk}", 360, ...)` writes the stale-derived chunk back with a fresh 360 s TTL. Now: Redis = {TOKEN' (new metadata) + chunk derived from old TOKEN}. there are internally inconsistent: metadata and chunks disagree, and nothing can tell. That's the dangerous zone! the user opens a dataset and gets mismatched metadata.

Same failure happens for plain TTL expiry: if Redis drops the keys but w2's `TTLCache` clock hasn't reached yet, w2 keeps serving and re-populating Redis from data that Redis has already expired.

I think the hotfix is to consult Redis before trusting the local hit and evict if it doesn't exist:

```suggestion
        result = self._cache_lookup(key)
        if result is not None:
            if self.cache.exists(key, f"{key}-dset") == 2:
                return result
            self._evict_object_cache(key)
```

tropicrainforest (commented):
One main comment on caching. there's an inconsistency between the per-worker cache and Redis. This could be dangerous. there's a scenario where the user asks for a dataset and gets back metadata that doesn't match the chunks, because the response is served from a stale per-worker cache under multi-worker mode without consulting Redis beforehand.

freva-data-portal-worker/src/data_portal_worker/load_data.py
antarcticrainforest: I wouldn't really call this a bug - it's a feature and a good one. This particularly TTL (not to be confused with the TTL of the sahred zarr url) is *not* a security feature,  it is for preventing the memory of the redis server from overflowing with stale data,. Any request for data to the rest-api triggers a new loading of cached data if it's not present anyway. So anyone can re-trigger loading cache anyway.

If the code hits the place you're concerned about then we have an empty redis cache but still data in the local cache, then we probably want to repopulate the data antway. If we wouldn't do that and we would continue running into non existing chunks xarray would end up with `NaN` arrays which is probably not what we want at all.

antarcticrainforest:
@tropicrainforest how are we going with this?

freva-data-portal-worker/data-loader
```diff
@@ -161,22 +161,17 @@ scheduler_port=${scheduler_port%%/*}
...
+passwd=$(base64 -d "${CONFIG_FILE}" | jq -r .passwd)
+cache_exp=$(base64 -d "${CONFIG_FILE}" | jq -r .cache_exp)
+TEMP_DIR=$(mktemp -d)
+${MAMBA_DIR}/bin/git clone https://github.com/freva-org/freva-nextgen.git $TEMP_DIR || { rm -fr $TEMP_DIR; exit 1; }
+${MAMBA_DIR}/bin/python -m pip install -U $TEMP_DIR/freva-data-portal-worker || { rm -r $TEMP_DIR; exit 1; }
+rm -rf $TEMP_DIR
+CMD="${MAMBA_DIR}/bin/python -m data_portal_worker \
+    -c ${CONFIG_FILE} -e \
+    $cache_exp -r $redis_host -p \
+    $scheduler_port"
```
tropicrainforest: let's keep this align with the newly defined `-p` flag, which is number of procs and drop`-p $scheduler_port"`. I'm not sure if it breaks the data-loader on run time or not

tropicrainforest (approved):
LGTM!

freva-data-portal-worker/src/data_portal_worker/cli.py
```diff
@@ -205,10 +264,15 @@ def run_data_loader(argv: Optional[List[str]] = None) -> None:
...
+        type=int,
+        default=max(1, min(8, int(os.getenv("API_NUM_PROCS", str(os.cpu_count()))))),
+    )
     args = parser.parse_args(argv)
     log_level = _set_loglevel_from_verbosity(args.verbose)
     kwargs = {
-        "port": args.port,
         "exp": args.exp,
         "redis_host": args.redis_host,
         "dev": args.dev,
```
tropicrainforest: in upstream `run_process` doesn't get `dev` arg anymore, yes or I'm wrong?

freva-data-portal-worker/src/data_portal_worker/load_data.py
```diff
@@ -523,16 +568,19 @@ def redis_callback(
         body: bytes,
     ) -> None:
         """Callback method to receive rabbit mq messages."""
+        data_logger.critical(body)
```
tropicrainforest: I believe it was a test log

tropicrainforest (approved):
this time really LGTM!

freva-data-portal-worker/data-loader
```diff
@@ -161,22 +161,16 @@ scheduler_port=${scheduler_port%%/*}
...
+user=$(base64 -d "${CONFIG_FILE}" | jq -r .user)
+passwd=$(base64 -d "${CONFIG_FILE}" | jq -r .passwd)
+cache_exp=$(base64 -d "${CONFIG_FILE}" | jq -r .cache_exp)
+TEMP_DIR=$(mktemp -d)
+${MAMBA_DIR}/bin/git clone https://github.com/freva-org/freva-nextgen.git $TEMP_DIR || { rm -fr $TEMP_DIR; exit 1; }
+${MAMBA_DIR}/bin/python -m pip install -U $TEMP_DIR/freva-data-portal-worker || { rm -r $TEMP_DIR; exit 1; }
+rm -rf $TEMP_DIR
+CMD="${MAMBA_DIR}/bin/python -m data_portal_worker \
+    -c ${CONFIG_FILE} -e \
+    $cache_exp -r $redis_host -p"
```
tropicrainforest: I think `-p` also should be removed

PR #364, opened 2026-05-27
Hotfix.

antarcticrainforest:
Some redis connection settings were broken, this tries to fix it.


docker-scripts/entrypoint.sh
```diff
@@ -1,74 +1,125 @@
...
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_USER '.user'
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_PASSWORD '.passwd'
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_HOST '.host'
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_SCHEDULER_HOST '.scheduler_host'
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_CACHE_EXP '.cache_exp'
```
tropicrainforest: On docker-based deployments, i think the REDIS cache expiry doesn't take effect because of the naming. but on conda-based deployments this isn't an issue, since the entrypoint isn't involved at all. because the real name is `API_CACHE_EXP`.

docker-scripts/entrypoint.sh
```diff
@@ -1,74 +1,125 @@
...
+
+    # Redis connection/config values. Missing, null, or empty values are unset.
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_USER '.user'
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_PASSWORD '.passwd'
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_HOST '.host'
+    export_json_value_if_not_empty "$decoded_cache_config" \
+        API_REDIS_SCHEDULER_HOST '.scheduler_host'
```
tropicrainforest: i believe it;s the leftover of last PR, not sure

tropicrainforest (approved):
LGTM!

docker-scripts/entrypoint.sh
antarcticrainforest: only sceduler host

docker-scripts/entrypoint.sh
antarcticrainforest: good catch it's called API_REDIS_CACHE_EXP

PR #365, opened 2026-05-29
Improved logging

antarcticrainforest:
Small fixes to ged rid of some logging noise and improve hot reloading


freva-rest/src/freva_rest/rest.py
```diff
@@ -98,8 +98,11 @@ async def refresh_extended_search_cache_periodically(
 
     while not stop_event.is_set():
         try:
-            for key in ("file", "uri"):
-                await Solr.refresh_extended_search_cache(uniq_key=key, max_results=100)
+            with set_logger_level("httpx", "httpcore", "freva-rest"):
+                for key in ("file", "uri"):
+                    await Solr.refresh_extended_search_cache(
```
tropicrainforest, line 106: not important but just as a question for my own learning: if we can generalize suppression to whole `freva-rest` we probably lose logs? One scenario: since the mute is open across the `await Solr.refresh_extended_search_cache(`, so i think it suppresses INFO logs as long as the refresh runs, which means, if Solr goes slow for any reason, that takes less than timeout seconds on freva-rest, and any user query running during it loses its freva-rest below WARNING logs too. Solr slow down -> don't know what user queried based on log. But since still we have them on the Mongo as stat, there is nothing to be worried about queries of users for pinpointing the user errors.

antarcticrainforest: Yeah I was thinking about the same. I've added a filter that sets log entries smaller than a given level (warning) to debug. The filter should user a [contextvar](https://docs.python.org/3/library/contextvars.html#module-contextvars) so that loggers outsied the context are not affected.

antarcticrainforest:
@tropicrainforest can you have a look again

freva-rest/src/freva_rest/logger.py
```diff
@@ -74,6 +61,43 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+        this = cls()
+        for name in names:
+            logging.getLogger(name).addFilter(this)
+        token = cls._ctx.set((level, frozenset(names)))
+        try:
+            yield
+        finally:
+            cls._ctx.reset(token)
+            for name in names:
+                logging.getLogger(name).removeFilter(this)
```
tropicrainforest: I tried to make sense of attaching/detaching the filter in the new names loop on enter and exit, and I might be wrong here. It doesn't break anything, but I think in overlap situations it silently stops doing the filtering we expect and misbehave.
For example, if `floor` were used on a concurrent path, for example wrapping extended_search, and 10 users hit it at once, they'd all share the one filter. The moment the first request finishes, it calls `removeFilter` and detaches it from the shared logger, even though the other 9 are still loitering inside this function. So the filter is gone, and for the rest of their run it's doing no filter and the winner destroys the bridge for the rest.
Nothing to worry about right now, becasue we don't use `floor` on the critical concurrent paths in core, specifically the most used one `extend_search` on web. But a docstring saying not to use it in concurrent tasks would be a nice safeguard. And since it only affects logs, even if a dev used it by mistake it wouldn't hurt much.

antarcticrainforest: Right, so I attach them in reset_looger and do not attache them anymore that should do it

tropicrainforest (approved):
LGTM!

antarcticrainforest:
Thanks!

PR #366, opened 2026-06-01
Hotfix

antarcticrainforest:
This PR avoids to caching a potentially empty solr facet list. The facet list is loaded on startup, if solr isn't ready the facet list will stay empty.

With this change the facet list is still empty but on the next user query the api tries to re query the facets from solr.  There was also a bug that produced a non empty facet list `['']` instead of `[]` so the facet query in the extended search endpoint would query empty facets instead of `all` -> facets `facets = config.solr_fields or "all"`.


eelucio:
LGTM (just in the whatsnew)


is it a pytest for these contigencies needed? (chatgpt suggestion),

1. for Solr unavailable at startup:
* A test where _get_solr_fields() fails during config startup, leaving solr_fields == []
* Then a later request re-accesses solr_fields, Solr is “back”, and the fields are fetched successfully
```python
def test_solr_fields_retry_after_startup_failure(
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    """An empty startup fetch should be retried on the next access."""
    ServerConfig._instance = None
    ServerConfig._initialised = False
    monkeypatch.setenv("API_TESTS", "1")

    calls = 0

    def fake_get_solr_fields(self: ServerConfig) -> list[str]:
        nonlocal calls
        calls += 1
        if calls == 1:
            return []
        return ["project", "experiment"]

    monkeypatch.setattr(ServerConfig, "_get_solr_fields", fake_get_solr_fields)

    cfg = ServerConfig()

    # First call happens during model_post_init()
    assert cfg._solr_fields == []

    # Second access should retry and refill the cache
    assert cfg.solr_fields == ["project", "experiment"]
    assert cfg._solr_fields == ["project", "experiment"]
    assert calls == 2
```

2. A regression test that ensures failure does not produce [""]
```python
def test_solr_fields_failure_stays_empty_list(
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    """A failed Solr field fetch should not create ['']."""
    ServerConfig._instance = None
    ServerConfig._initialised = False
    monkeypatch.setenv("API_TESTS", "1")

    monkeypatch.setattr(ServerConfig, "_get_solr_fields", lambda self: [])

    cfg = ServerConfig()

    assert cfg.solr_fields == []
    assert cfg.solr_fields != [""]
```

antarcticrainforest:
Yeah, agreed we could add some tests.

freva-rest/src/freva_rest/config.py
```diff
@@ -574,10 +573,11 @@ def set_debug(debug: bool) -> None:
         logger.setLevel(level)
         logger_file_handle.setLevel(level)
 
-    @cached_property
+    @property
     def solr_fields(self) -> List[str]:
         """Get all relevant solr facet fields."""
-        return list(self._solr_fields)
+        self._solr_fields = self._solr_fields or self._get_solr_fields()
+        return self._solr_fields
```
tropicrainforest, line 586: The docker CI build fails because the Solr facets are empty at startup, and building `ServerConfig()` in the CLI triggers a Solr web request immediately. I think the fix is to only fetch the facets when they're actually needed, by checking that `_solr_fields` is not `None` first. checked on local and it seems to be working

antarcticrainforest: Nah, I'd like to fetch them on startup. Then we don't have to wast any time when user requests coming in.

antarcticrainforest:
Ok the borken, tests are working again. @eelucio and @tropicrainforest ready to rumble?

tropicrainforest (approved):
LGTM!

PR #371, opened 2026-07-01
Use backend_kwargs to propagate decode options

mannreis:
A workaround for https://github.com/freva-org/xarray-prism/issues/6


antarcticrainforest:
@tropicrainforest , can you look at this?

tropicrainforest:
@mannreis since CI is still red, what is the latest status here? Is it already ready for review?

tropicrainforest (commented):
@antarcticrainforest this issue has been fixed upstream and in the next release of freva-data-loader, the change would be applied when it builds the new image or conda version. So I think we can close this PR

PR #372, opened 2026-07-14
feat: let extended-search return arbitrary result fields

tropicrainforest:
The new databrowser web UI needs `bbox` and `time` visible in the results, so this PR adds a `fields` query parameter to the extended-search endpoint ` /databrowser/extended-search/freva/uri?fields=time&fields=bbox`
Also it can be generalised to any field in the managed schema so it's not only limited to bbox and time. To keep the API from buckling under the back pressure, the number of requested fields is capped at 25. Without `fields`, the response is unchanged to `['uri', 'fs_type']` to keep it backward compatible.

Also this PR includes a couple of flavour fixes that came out of this work.

https://freva-deployment.dkrz.de/api/freva-nextgen/databrowser/extended-search/freva/file?file=s3://icdc/healpix/atmosphere/ASCAT/METOP_B/P1D/level_6.zarr&fields=time&fields=bbox


antarcticrainforest (approved):
I see some gaps with freely chosen user flavours and this evaluation. Might be worth documenting. But we can do that if it actually becomes a problem:

LGTM

